Issue #18, opened 2025-09-18
request: add `permissions` to all workflows


mdhaber:
### Motivation

It is generally recommended that all workflows should contain explicit `permissions`s rather than relying on repository defaults.

### Proposed solution

Add `permissions` to all workflow `yml`s.

PR #30, opened 2025-09-19
tests: add unit tests of _logicals.py

mdhaber:
## 📝 Description

Adds unit tests of `deltakit-explorer/tests/unit/codes/test_logicals.py::paulistring_to_operator`.

## 🧾 Release Note

PR title.


nelimee (approved):

nelimee:
Permissions to CI jobs should be added (see #18) in order for CI to pass here.

anjbur:
Looks like the CI failure is because of permissions issues arising from the source branch being in a fork. This is a bit tricky to fix securely, as it will require splitting the current `labeled_tests.yml` workflow into two: one which has write permissions and only labels code, not checking anything out; and a second that reverts to standard fork read-only permissions and checks out code to run tests. Until then, the CI will create this failure on any PR from a branch in a fork.

I've created an issue (#49) to track this and will look further into implementing the fix next week. If we need to unblock fork PRs before then, we could add a temporary check that just runs all tests when the source branch is in a fork, instead of trying to label what's changed.

PR #47, opened 2025-09-19
feat(cancel): cancelling the job with Ctrl+C in APIv2

str-anger:
## 🔗 Closed Issues

A bug from backlog 

---

## 📝 Description

This PR re-implements an approach already present in APIv1.
If the request is sent, we may cancel it by CTRL+C. In this case,
`KeyboardInterrupt` exception is raised. We handle this and try to
cancel the job on the client side. Then we replace exception with a
meaningful message.

---

## 🚦 Status

Done.

---

## 🛠️ Future Work

Follows by a PR on the server side.

---

## ➕️ Additional Information

N/A

---

## 🧾 Release Note

PR Title

### DIFF
--- a/deltakit-explorer/src/deltakit_explorer/_api/_api_v2_client.py
+++ b/deltakit-explorer/src/deltakit_explorer/_api/_api_v2_client.py
@@ -206,7 +206,7 @@ class APIv2Client(APIClient):
             return job.result
         except KeyboardInterrupt:
             count = self.kill(job.request_id)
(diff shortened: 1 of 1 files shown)


str-anger:
@nelimee thank you for the link. This is a very valid comments, we already explored this place in our previous project.
1. We are not catching it silently, we are catching and failing. I think we can tolerate failed stated as we've already `SIGINT` the program.
2. Proposed example is (to me) an overengineering. We had it in a server software for exactly the reason explained. I don't feel we should be very careful in a client lib.

PR #59, opened 2025-09-22
ci: Fix CI for fork PRs

### LINKED ISSUE #49: request: refactor the labeled_tests.yml workflow to work on PRs from forks
### Motivation

The `labeled_tests.yml` workflow fails on PRs from forks, as such PRs don't get write access by default.

### Proposed solution

To securely fix this, we'll need to split `labeled_tests.yml` into two workflows:
- The first will only label tests without checking out any code, using the `pull_request_target` trigger to have the correct permissions. As written, we currently check out the code to access a python script that's run against the labels. This will need to be refactored into the workflow itself, so that we aren't checking out and running code from a fork at full write permissions with secret access.
- The second workflow can have a `workflow_run` trigger, so that it runs after the labeler completes.

### Potential alternatives

_No response_

### Additional context

_No response_

anjbur:
## 🔗 Closed Issues

Closes #49

---

## 📝 Description

- Separate labeling and testing into separate workflows. This allows the labeling action to safely have write permission and restricts the test action to read permission only.
- Label PRs with `dev` if they change files in `.github/` or `tools/`

---

## 🧾 Release Note

PR title

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@
 [![SPEC 0][spec0-badge]][spec0-link]
 [![Issues][issues-badge]][issues-link]
 [![Discussions][discussions-badge]][discussions-link]
+[![DOI][doi-badge]][doi-link]
 
 [docs-badge]: https://readthedocs.org/projects/deltakit/badge/?version=latest
 [docs-link]: https://deltakit.readthedocs.io/en/latest/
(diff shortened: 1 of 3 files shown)


nelimee (commented):
It seems fine to me, but I think another review would be nice as I am not sure to understand all the changes.

mdhaber:
Looks fine. `dev` label looks good. Let's give it a shot!

PR #65, opened 2025-09-25
fix: pre-commit hook configuration and fixes

### LINKED ISSUE #63: bug: enforce pre-commit rules
### Bug Description

Currently, `pre-commit run -a` (i.e., run `pre-commit` on all the files, independently of their commit/modification status) fails and does not produce the expected output. A non-exhaustive list of the issues:

- `isort` is run on all packages, but `deltakit-core` and `deltakit-circuit` imports are managed by `ruff format` now. 
- pre-commit rule `name-tests-test` fails on an utility sub-module in tests.
- `ruff check` fails on `deltakit` tests and tools.
- `mypy` rule fails.
- `bandit` rule fails.

### Minimal Reproducible Example (MRE)

```sh
git checkout main
pixi run pre-commit run -a
```

### Actual behavior

_No response_

### Expected behavior

Running `pre-commit run -a` on `main` should not modify any file and should succeed.

### Additional context

_No response_

nelimee:
## 🔗 Closed Issues

Fixes #63

---

## 📝 Description

This PR ensures that `pre-commit` hooks are correctly configured and that `pre-commit run -a` passes without any fail.

---

## 🚦 Status

Some work needs to be done to pass `pre-commit run -a` without any fail.

---

## 🛠️ Future Work

None

---

## ➕️ Additional Information

None

---

## 🧾 Release Note

PR title


.pre-commit-config.yaml
```diff
@@ -2,14 +2,14 @@ repos:
...
-      - id: check-yaml
-      - id: end-of-file-fixer
       - id: check-added-large-files
       - id: check-case-conflict
       - id: check-merge-conflict
       - id: check-symlinks
       - id: check-yaml
+      - id: check-toml
       - id: debug-statements
+      - id: detect-private-key
```
mdhaber, line 12: Ooh, nice addition

.pre-commit-config.yaml
```diff
@@ -55,9 +48,3 @@ repos:
     rev: v1.7.7
     hooks:
       - id: actionlint
-
-  - repo: https://github.com/pycqa/isort
-    rev: 6.0.1
-    hooks:
-      - id: isort
-        name: isort (python)
```
mdhaber, line 63: The goal is to have `ruff format` take care of `isort` rules, and the two conflict. So removing this.

mdhaber (approved):
Thank you!

At one point, these were all working except for `mypy`, but I started getting false positives from `mypy` so much (and thought it was just a me problem) that I started using `--no-verify`. So that's how these other things must have slipped in without notice.

mdhaber:
Looks like RTD is held up by concurrency limit. I'll merge this when it finishes.
In the meantime, I'll get `ruff format` running in pre-commit.

Issue #66, opened 2025-09-25
request: error budgeting


nelimee:
### Motivation

*Note: this issue will be edited later on to include more details*

-------------------

It would be nice to have an easy way of producing error-budgets for different noise models.

### Proposed solution

Implement error-budgeting according to the methodology outlined in several research papers.

PR #67, opened 2025-09-25
feat: better logical error probability per round

nelimee:
## 🔗 Closed Issues

Towards #66

---

## 📝 Description

Improve the numerical methods used to compute the logical error probability per round from the raw logical error probabilities obtained through experiments.

---

## 🚦 Status

Ready to merge.

---

## 🛠️ Future Work

- Improve numerical methods used to compute lambda.
- Implement new methods to have a full pipeline to compute the error-budget for a given noise model.

---

## ➕️ Additional Information

This PR has already been reviewed by several people internally. More reviews are welcomed.

---

## 🧾 Release Note

PR title.


deltakit-explorer/src/deltakit_explorer/analysis/_analysis.py
```diff
@@ -192,17 +564,20 @@ def calculate_lambda_and_lambda_stddev(
...
     """
     query_id = Logging.info_and_generate_uid(locals())
-    if min(distances) < 5:
-        Logging.warn("Lambda unreliable at low code distances."
-                        "Please use distance 5 as a minimum.", query_id)
+    # Make sure that the inputs are numpy arrays sorted by distance
+    isort = np.argsort(distances)
+    distances = np.asarray(distances)[isort]
+    lep_per_round = np.asarray(lep_per_round)[isort]
+    lep_stddev_per_round = np.asarray(lep_stddev_per_round)[isort]
```
CaiReesManuelRL, line 577: I think there is an implicit assumption here that we are using odd distances, which I believe is required to fit to `Lambda^(-(d + 1) / 2)`. I think it makes sense to document and enforce this somewhere. 

I don't know if there is a way to generalise the fitting to even distances. Naively `Lambda^(-floor((d + 1) / 2))`, where `floor((d + 1) / 2)` is the shortest un-decodable error chain for both even and odd distance codes, but I think this says that a odd distance code `d` and even `d+1` should have the same error rate per round which is likely not true.

CaiReesManuelRL (approved):
Great work, src code looks good to me, have left some minor comments

deltakit-explorer/src/deltakit_explorer/analysis/_analysis.py
nelimee: In fact, already addressed in #68 with
```py
    # Make sure that there are no even distances.
    if np.any(distances % 2 == 0):
        raise ValueError(
            "Found at least one even distance in the provided distances "
            f"({distances.tolist()}). This is not supported."
        )
```

nelimee:
A lot of eyes went on this code, I think this is ready to be merged.

guenp (approved):
Great work, thank you for the extensive docstrings and inline comments.

Issue #27, opened 2025-09-18
request: consider enforcing additional lint rules


mdhaber:
### Motivation

We use `ruff check` with many of the most important lint rules, but there are some others that we should consider.

### Proposed solution

Consider implementing all the rules from the Scientific Python [Style and static checks](https://learn.scientific-python.org/development/guides/style/#ruff) guide.

- [ ]  "ARG",      # flake8-unused-arguments
- [ ]  "B",        # flake8-bugbear
- [ ]  "C4",       # flake8-comprehensions
- [ ]  "EM",       # flake8-errmsg
- [ ]  "EXE",      # flake8-executable
- [ ]  "FURB",     # refurb
- [ ]  "G",        # flake8-logging-format
- [ ]  "I",        # isort
- [ ]  "ICN",      # flake8-import-conventions
- [ ]  "NPY",      # NumPy specific rules
- [ ]  "PD",       # pandas-vet
- [ ]  "PGH",      # pygrep-hooks
- [ ]  "PIE",      # flake8-pie
- [ ]  "PL",       # pylint
- [ ]  "PT",       # flake8-pytest-style
- [ ]  "PTH",      # flake8-use-pathlib
- [ ]  "PYI",      # flake8-pyi
- [ ]  "RET",      # flake8-return
- [ ]  "RUF",      # Ruff-specific
- [ ]  "SIM",      # flake8-simplify
- [ ]  "T20",      # flake8-print
- [ ]  "UP",       # pyupgrade
- [ ]  "YTT",      # flake8-2020

PR #78, opened 2025-09-26
style: enforce NumPy-specific lint rules

zhiyaol:
## 🔗 Closed Issues

Toward #27 

---

## 📝 Description

This PR adds the NPY (NumPy-specific) rules by editing the ruff configuration in ruff.toml and fixes a lint rule violation by replacing legacy `np.random.uniform` call with `np.random.Generator`

---

## 🚦 Status

Ready for review.

---

## 🧾 Release Note

PR title

### DIFF
--- a/.github/workflows/labeled_tests.yml
+++ b/.github/workflows/labeled_tests.yml
@@ -58,5 +58,4 @@ jobs:
         run: pixi add python=${{ matrix.python-version.version }}
 
       - name: Run ${{ matrix.package }} tests
-        if: contains(needs.labels-to-json.outputs.all-labels, matrix.package)
         run: pixi run -e ${{ matrix.package }}-tests-py${{ matrix.python-version.label }} tests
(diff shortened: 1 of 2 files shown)


mdhaber (approved):
Thanks @zhiyaol! This looks good. 

One thing that is worth mentioning is that this test is nondeterministic because `default_rng` is not provided with a seed. Nondeterministic tests can be very useful to find bugs, but it is sometimes considered good practice to make testing in CI deterministic.

Consider what happens when there is a bug that causes the test to fail only sometimes, depending on the random number that is generated. If this nondeterministic test is in `main` and someone is working on something else, they might just happen to experience an error in the test. At the least, that's distracting from their work; worse, they could spend some time _thinking_ it's related to their work and trying to debug something that has nothing to do with their work.

Instead, "nondeterministic" testing can be performed offline to find these sorts of bugs. I put "nondeterministic" in quotes because it would be done in a way that makes the failures reproducible, and it's done by someone interested in finding and fixing errors in _that_ code. [Hypothesis](https://hypothesis.readthedocs.io/en/latest/) is a package that supports this sort of testing.

**That said**, this test was nondeterministic to begin with because the global random seed was not set, so I think it's OK to preserve that. If we want to set a project-wide policy in favor of or against nondeterministic testing, that can be done separately from this PR.

Thanks again!

Issue #14, opened 2025-09-18
request: read the docs integration improvements


mdhaber:
### Motivation

In gh-9, the notebooks aren't running correctly because RTD can't find the token for using the cloud API. This was only discovered manually; RTD reported successful build.

### Proposed solution

- RTD needs to be configured with a token to use the cloud API
- RTD should report a failure status if there is an error in the docs build.

PR #86, opened 2025-10-02
chore: simplify .readthedocs.yaml; add `-W` (fail on warnings) to custom build command

mdhaber:
Closes gh-14

## 📝 Description

Read the Docs is not respecting the `fail_on_warning: true` setting because we're using a custom doc build command. The simplest fix is to add `-W` in the custom build command, but I'd like to see if I can remove the custom build command.

---

## 🚦 Status

Draft/testing

## 🧾 Release Note

PR title

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,11 +9,9 @@ build:
   os: ubuntu-24.04
   tools:
     python: "3.13"
-
-python:
-  install:
-    - method: pip
-      path: .[docs]
+  commands:
+    - curl -sSf https://pixi.sh/install.sh | sh
+    - /home/docs/.pixi/bin/pixi run sphinx-build -T -W -n -b html -d _build/doctrees -D language=en docs $READTHEDOCS_OUTPUT/html
 
 # Build documentation in the "docs/" directory with Sphinx
 sphinx:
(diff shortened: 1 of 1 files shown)


.readthedocs.yaml
```diff
@@ -9,9 +9,11 @@ build:
...
   tools:
     python: "3.13"
-  commands:
-    - curl -sSf https://pixi.sh/install.sh | sh
-    - /home/docs/.pixi/bin/pixi run sphinx-build -T -b html -d _build/doctrees -D language=en docs $READTHEDOCS_OUTPUT/html
+
+python:
+  install:
+    - method: pip
+      path: .[docs]
```
mdhaber, line 16:
```suggestion
  commands:
    - curl -sSf https://pixi.sh/install.sh | sh
    - /home/docs/.pixi/bin/pixi run sphinx-build -T -W -n -b html -d _build/doctrees -D language=en docs $READTHEDOCS_OUTPUT/html
```

mdhaber:
I guess RTD ignores the whole `sphinx` section of the `.readthedocs.yml` when there is a custom build command. Makes sense. So this PR removes the ignored section and adds `-W` to the custom build command. It now shows a failure here when there are warnings. Hopefully this CI run passes, but the failure report can be confirmed by clicking the red X next to `ee72094`.

image

PR #87, opened 2025-10-03
feat: add plotting functionality to `compute_logical_error_per_round`

zhiyaol:
## 🔗 Closed Issues

Toward #66 
---

## 📝 Description

Adding the option to plot the exponential curve that estimated LEPPR in `compute_logical_error_per_round`

---

## 🚦 Status

1. The `estimated_logical_error_per_round_stddev` calculated from example seems quite big. 
2. Attempted to also plot the error band for the exponential curve, but the propagated errors I calculated are very big, which is a similar issue to 1. The very wide error band obscure features of the exponential curve.

---

## 🛠️ Future Work

Proper error propagation.

---

## ➕️ Additional Information

Will discuss with @nelimee.

---

## 🧾 Release Note

PR title.

### DIFF
--- a/deltakit-explorer/src/deltakit_explorer/analysis/_analysis.py
+++ b/deltakit-explorer/src/deltakit_explorer/analysis/_analysis.py
@@ -12,6 +12,9 @@ import numpy.typing as npt
 from scipy.optimize import curve_fit
 from deltakit_explorer._utils._logging import Logging
 
+import matplotlib.pyplot as plt
+colors = ["#006F62", "#FF7500", "#000000"]
+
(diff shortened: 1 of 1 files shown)


nelimee (commented):
I would largely prefer the visualisation code to be:

1. in a separate, isolated, function whose sole purpose is to plot, that helps in having simpler and shorter functions,
2. in a separate module, so that users that only want to compute the logical error-rate per round do not have to import `matplotlib`.

If you have a plot showing the standard deviations that seems too large (and the code generating it), that would be nice.

Issue #61, opened 2025-09-22
request: Contributing guide for Conda or Miniconda


guenp:
Currently, it's not clear to me that it's possible to use `pixi` within a Miniconda environment. After installing Pixi on a machine with Miniconda installed and creating a new environment using `pixi install` and activating it with `pixi shell`, it seems to be using the Python installation in the Miniconda `base` environment.

We should either 1) figure out how to make Conda work with `pixi`, or 2) add instructions to the contribution guide to include instructions for how to set up a development environment with Conda.

Repro steps:
1. Install Miniconda
2. Install Pixi
3. Run `pixi install`
4. Run `pixi shell`
5. Importing deltakit results in "package not found" error.

CC @zhiyaol

mdhaber:
@zhiyaol is going to write a note in `CONTRIBUTING.md` addressing the conda issue. The main problem was that conda was automatically activating an environment whenever a terminal was opened - and even after deactivating the environment it, it was re-activated by `pixi shell`. There's a one-liner to turn this off. Conda and pixi can coexist on the same machine, but it's not advised to use both to manage environments simultaneously.

nelimee:
> [@nelimee](https://github.com/nelimee) Subtle distinction, I think:
> 
>   * [request: test with`poetry`-installed `deltakit` #37](https://github.com/Deltakit/deltakit/issues/37) was about _running tests_ with (non-editable) `deltakit` installed via `poetry` - or, per your comment, uv, pip, etc.
>   * This issue is about setting up a _development environment_ with `pixi` on a machine that has Conda / Miniconda installed.
>   * There is a third issue we might have, which would be to support setting up development environments using a tool other than `pixi`. Currently [@zhiyaol](https://github.com/zhiyaol)'s development is in set up with Conda; not `pixi`.

You are right. On my side, I think the last point is important. Having to use an external tool you are not familiar with to setup a dev environment has always been a hassle for me with new repositories, and that is a friction I would like to remove completely from the new contributor journey.

PR #92, opened 2025-10-03
dev: add support for package managers other than `pixi`

mdhaber:
## 🔗 Closed Issues

Closes gh-61
Closes gh-92

---

## 📝 Description

This adds support for managing development environments with Conda, Poetry, Hatch, and uv.  After installing the desired package manager, please navigate to the deltakit directory, deactivate any other virtual environment / shell, and follow the instruction in the documentation to test the package manager.

Note that users of these other tools won't be able to use pixi's task runner, but it is still easy to use the important tools (pre-commit, pytest, ruff, mypy, etc.) directly. These all work for me:

- `pytest deltakit-explorer` (see gh-94, which makes `pytest` run tests of all component packages)
- `mypy deltakit-explorer`
- `ruff check`
- `ruff format deltakit-circuit`
- `typos`
- `pre-commit run -a`

If you don't have a virtual environment / shell activated with the chosen tool, you'll need a prefix like `poetry run` or `uv run` before the commands.

It would also be good to confirm the environment does indeed do an _editable_ install of Deltakit, e.g.:
- make a change to a Deltakit feature that is easy to test. (This should be an edit to a deltakit component package, like files in `deltakit-explorer`, not just in top-level `deltakit`. I've found that it's easy to get top-level deltakit installed in editable mode, but then component packages may still be installed from `pip`, or from local source but not editable!)
- start the package manager's version of `python`
- use the changed feature feature to see that the change is reflected.

Personally, I've been testing this by adding and remove `raise ValueError('hi')` in the `deltakit-explorer` `__init__.py` and checking that I do or don't get the error, as expected, when I `import deltakit.explorer`.

Note that users of tools other than `pixi` and `conda` may need to install `pandoc` separately if they want to build documentation locally.

---

## 🚦 Status

- [x] Need to fix tabs in CONTRIBUTING.md
- [x] Need to fix interference between `uv` and `pixi`
- [x] `hatch` is not installing component packages in editable mode
- [x] `hatch` is not even installing component packages from local source
- [x] `uv` is not installing component packages in editable mode
- [x] `uv` is not even installing component packages from local source

---

## 🛠️ Future Work

- [x] Consider documenting
- [x] Test creating dev environments with `uv`, `pip`

---

## 🧾 Release Note

PR title.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,6 +89,7 @@ Depending on system settings, you may experience a `Too many open files (os erro
 error. This is [known issue](https://github.com/prefix-dev/pixi/issues/2626) that can easily be
 resolved by increasing the maximum number of open file descriptors; e.g., `ulimit -n 512`.
 ```
+
 ```{dropdown} Conda users...
 We suggest deactivating any `conda` environments before using `pixi`. If the `conda` `base` environment activates by default whenever a terminal session is opened, you can turn it off with `conda config --set auto_activate_base false`.
 ```


environment.yml
```diff
@@ -0,0 +1,16 @@
+name: deltakit-conda
+channels:
+- conda-forge
+- nodefaults
+dependencies:
+- pandoc >=3.7.0.2,<4
+- python >=3.10
+- actionlint >=1.7
+- pixi-pycharm >=0.0.8,<0.0.9
```
nelimee, line 9:
```suggestion
- python >=3.10
```
wouldn't the other be installed at least with the `-e ./[dev]`?

.gitignore
```diff
@@ -105,8 +105,8 @@ ipython_config.py
 #   This is especially recommended for binary packages to ensure reproducibility, and is more
 #   commonly ignored for libraries.
 #   https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control
-#poetry.lock
-#poetry.toml
+poetry.lock
+poetry.toml
```
nelimee, line 109: Is this ignore wanted?

nelimee (commented):
> `conda env create -f environment.yml`

only took a few seconds on my machine. 
I checked all the points for `conda`, everything seems to work as expected.

`poetry` fails at installation (`poetry install --extras "dev"` in the main `deltakit` folder) with
```
('Configuration file exists at /Users/suau/Library/Application Support/pypoetry, reusing this directory.\n\nConsider moving TOML configuration files to /Users/suau/Library/Preferences/pypoetry, as support for the legacy directory will be removed in an upcoming release.',)

The Poetry configuration is invalid:
  - 'name' is a required property
  - 'version' is a required property
  - 'description' is a required property
  - 'authors' is a required property
```

seems like a configuration file is missing.

environment.yml
mdhaber: No, as things are, none of these are installed with `-e ./[dev]`. You can tell because they are not listed in `pyproject.toml`.

We might be able to install `actionlint` via `actionlint-py`, but that is not what we currently have. 

We can remove `pixi-pycharm`, since that would not be needed for conda users, but the rest are needed for some development activities.

.gitignore
mdhaber: It was intended. 
We are not currently committing lock files.
We could consider adding a `poetry.toml`, but I didn't need it.
What version of poetry do you have?

mdhaber:
> seems like a configuration file is missing.

What version of poetry do you have? I am using 2.2.1. That could potentially explain the difference in behavior. `poetry` is getting all the information it needs from `pyproject.toml` on my machine. Please also be sure to deactivate any other environments and run the command from the project folder.  The message suggests that it's looking for a configuration file somewhere else... the current folder should be the default, but you could use the `--directory=.` option to make it explicit.

environment.yml
mdhaber: Done. Leaving `actionlint` and `pandoc` alone for now.

mdhaber:
OK, I think `hatch`, `uv`, and the documentation are also ready for a spin.
image
See the dropdown at the bottom of [Development Environment and Common Tasks](https://deltakit--92.org.readthedocs.build/en/92/CONTRIBUTING.html#development-environment-and-common-tasks).

nelimee (commented):
Blindly following the `CONTRIBUTING.md`:

- `pixi` works fine and makes an editable install.
- `conda` works fine and makes an editable install.
- `poetry` fails (see below).
- `uv` works fine and makes an editable install.
- `hatch` works fine and makes an editable install.

------------------

On `poetry` fail, I have no idea where it comes from. It seems like poetry believes I am asking for `python>=3.14`, even though 
```toml
[tool.poetry.dependencies]
python = ">=3.10,<4.0"
deltakit-core = { path = "./deltakit-core/", develop = true }
deltakit-circuit = { path = "./deltakit-circuit/", develop = true }
deltakit-decode = { path = "./deltakit-decode/", develop = true }
deltakit-explorer = { path = "./deltakit-explorer/", develop = true }
```
is in the `pyproject.toml`. Below is the full error I get with `poetry install --extras dev`:
```
Updating dependencies
Resolving dependencies... (5.9s)

The current project's supported Python range (>=3.14,<4.0) is not compatible with some of the required packages Python requirement:
  - plotly-stubs requires Python <3.14,>=3.10, so it will not be installable for Python >=3.14,<4.0
  - plotly-stubs requires Python <3.14,>=3.10, so it will not be installable for Python >=3.14,<4.0
  - plotly-stubs requires Python <3.14,>=3.10, so it will not be installable for Python >=3.14,<4.0
  - plotly-stubs requires Python <3.14,>=3.10, so it will not be installable for Python >=3.14,<4.0
  - plotly-stubs requires Python <3.14,>=3.10, so it will not be installable for Python >=3.14,<4.0
  - plotly-stubs requires Python <3.14,>=3.10, so it will not be installable for Python >=3.14,<4.0

Because no versions of plotly-stubs match <0.0.1 || >0.0.1,<0.0.2 || >0.0.2,<0.0.3 || >0.0.3,<0.0.4 || >0.0.4,<0.0.5 || >0.0.5,<0.0.6 || >0.0.6
 and plotly-stubs (0.0.1) requires Python <3.14,>=3.10, plotly-stubs is forbidden.
And because plotly-stubs (0.0.2) requires Python <3.14,>=3.10, plotly-stubs is forbidden.
And because plotly-stubs (0.0.3) requires Python <3.14,>=3.10
 and plotly-stubs (0.0.4) requires Python <3.14,>=3.10, plotly-stubs is forbidden.
And because plotly-stubs (0.0.5) requires Python <3.14,>=3.10
 and plotly-stubs (0.0.6) requires Python <3.14,>=3.10, plotly-stubs is forbidden.
So, because deltakit depends on deltakit-decode (0.5.1) @ file:///Users/suau/Code/deltakit/deltakit-decode which depends on plotly-stubs (*), version solving failed.

  * Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties

    For plotly-stubs, a possible solution would be to set the `python` property to "<empty>"
For plotly-stubs, a possible solution would be to set the `python` property to "<empty>"
For plotly-stubs, a possible solution would be to set the `python` property to "<empty>"
For plotly-stubs, a possible solution would be to set the `python` property to "<empty>"
For plotly-stubs, a possible solution would be to set the `python` property to "<empty>"
For plotly-stubs, a possible solution would be to set the `python` property to "<empty>"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers
```

mdhaber:
Wow, great that so much went well. 
What changed since your last try with poetry to get such a different error? I can't imagine I did anything that resolved the original problem (if it were really as poetry described it). 
Would be great to get some others to try poetry. If it really is a config problem, maybe others will get errors with more clues.
I will try it on other machines - can you? How about a fresh poetry install, if you're willing?

nelimee:
I will try to explore more tomorrow.

PR #94, opened 2025-10-03
tests: allow `pytest` to run from top level

mdhaber:
## 📝 Description

`pytest` / `pixi run pytest` did not run properly from the top level of the repository. This fixes it so all tests in all packages run.

---

## 🚦 Status

Ready to review

---

## 🧾 Release Note

PR title

### DIFF
--- a/deltakit-decode/tests/__init__.py
+++ b/deltakit-decode/tests/__init__.py
@@ -1,1 +1,0 @@
-# (c) Copyright Riverlane 2020-2025.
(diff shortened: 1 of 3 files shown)


deltakit-decode/tests/__init__.py
mdhaber, line 1: This is really the file that happened to be causing trouble; the others just seemed unnecessary, too.

nelimee (approved):
Ran `pixi run pytest .` locally to double-check, no failure.

Issue #29, opened 2025-09-18
request: enforce additional typing rules


mdhaber:
### Motivation

`mypy` is performing some checks, but there are many failures that are currently not checked for or ignored. We need to work to eliminate these waivers over time.

All improvements welcome. To find current exclusions/ignores, see [`mypy.ini`](https://github.com/Deltakit/deltakit/blob/main/mypy.ini) and search code for `# type: ignore`.

### Proposed solution

[This](https://careers.wolt.com/en/blog/tech/professional-grade-mypy-configuration) looked like a good article about typing priorities, so with that in mind, these seem like the most important errors to fix:
- [ ] `disallow_untyped_defs`
- [ ] `disallow _any_unimported`
- [ ] `check_untyped_defs`
- [ ] `warn_return_any`
- [ ] `warn_unused_ignores`
- [ ] deltakit_circuit `disable_error_code`
- [ ] deltakit_decode`disable_error_code`
- [ ] deltakit_explorer `disable_error_code`, `follow-imports`

Lower priority:
- [ ] `disallow_incomplete_defs`
- [ ] `implicit_reexport`
- [ ] `warn_unreachable`
- [ ] `disallow_any_explicit`
- [ ] `disallow_any_expr`
- [ ] `disallow_any_decorated`
- [ ] address deltakit-circuit tests

rolandriver:
Addressing this issue now will incur work on possibly deprecated packages in the future. This will be revisited once upcoming packages are released.

PR #95, opened 2025-10-04
tests: allow mypy to run from top level 

mdhaber:
Toward gh-29.

## 📝 Description

`mypy .` did not run properly from the top-level directory. This fixes that so `mypy .` analyzes all component packages together.

---

## 🚦 Status

Ready for review

---

## ➕️ Additional Information

With the first commit `mypy .` fails with `407 errors in 20 files (checked 170 source files)`.  @nelimee This might have been what you were seeing when `pre-commit run -a` was doing a mypy check? When `mypy` runs from the top, I think it's able to spot problems involving interactions between packages.

I'm addressing this with a commit that adds the minimal set of ignores to silence the new errors. However, during the process I also found that with few modifications, we can also enforce more of the `strict` rules, and I was able to install stubs for a few more packages. Overall, I think this is a net win.

---

## 🧾 Release Note

PR title

### DIFF
--- a/deltakit-explorer/pyproject.toml
+++ b/deltakit-explorer/pyproject.toml
@@ -45,6 +45,7 @@ dependencies = [
     "pandas >= 2.0",
     "pandas-stubs",
     "types-seaborn",
+    "scipy-stubs",
 ]
 
 [project.urls]


.pre-commit-config.yaml
```diff
@@ -26,6 +26,33 @@ repos:
       # Run the formatter.
       - id: ruff-format
 
+  - repo: https://github.com/pre-commit/mirrors-mypy
```
mdhaber, line 29: With this, `pre-commit run -a` includes mypy checks and passes for me.

nelimee (approved):
The following pass without any error:
- `pixi run mypy .`
- `pixi run pre-commit run -a`
- `pixi run mypy-all`

Note that `pre-commit` is quite slow due to `mypy` on the first run, but subsequent run are way quicker once `mypy` cache is created, so that is fine.

PR #105, opened 2025-10-08
DEV: add upper bound on supported Python

mdhaber:
## 📝 Description

The Python 3.14 release is causing CI failures because there are some places where our Python version is unconstrained. When Python 3.14 gets installed before other packages, some other packages can't be installed.  This fixes the problem by adding an upper bound to Python version where needed.

---

## 🚦 Status

Ready for review.
---

## 🧾 Release Note

PR title

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -31,7 +31,6 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.13'
 
       - name: Set up Pixi
         uses: prefix-dev/setup-pixi@v0.8.10
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -29,9 +29,9 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up Pixi
-        uses: prefix-dev/setup-pixi@v0.8.10
+        uses: prefix-dev/setup-pixi@v0.9.1
         with:
-          pixi-version: v0.50.2
+          pixi-version: v0.56.0
 
       - name: Install dependencies
         run: |
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -28,11 +28,6 @@ jobs:
       - name: Checkout repo
         uses: actions/checkout@v4
 
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.x'
-
       - name: Set up Pixi
         uses: prefix-dev/setup-pixi@v0.8.10
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/docs.yml
```diff
@@ -31,7 +31,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.x'
+          python-version: '3.13'
```
mdhaber, line 34:
```suggestion
```

.github/workflows/docs.yml
```diff
@@ -31,7 +31,6 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
```
mdhaber, line 33:
```suggestion
        with:
          python-version: '3.x'
```

.github/workflows/docs.yml
```diff
@@ -28,15 +28,10 @@ jobs:
...
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.x'
-
       - name: Set up Pixi
-        uses: prefix-dev/setup-pixi@v0.8.10
+        uses: prefix-dev/setup-pixi@v0.9.1
         with:
-          pixi-version: v0.50.2
+          pixi-version: v0.56.0
```
mdhaber, line 34: Might as well consistently use the latest version.

pixi.toml
```diff
@@ -116,7 +116,7 @@ cmd = "isort ."
 
 [dependencies]
 pandoc = ">=3.7.0.2,<4"
-python = ">=3.10"
+python = ">=3.10,<3.14"
```
mdhaber, line 119: `pixi` resolves this before PyPI dependencies. So if there is no upper bound, `pixi` has no reason not to install Python 3.14. Then, when it's time to install the PyPI dependencies, there seems to be one that doesn't work with Python 3.14.

.github/workflows/docs.yml
nelimee: It is unclear to me if `latest` would work as a version requirement here. It seems like [`pixi-url` accepts such a version requirement](https://github.com/prefix-dev/setup-pixi?tab=readme-ov-file#download-pixi-from-a-custom-url), so maybe?

.readthedocs.yaml
```diff
@@ -7,8 +7,6 @@ version: 2
 # Set the OS, Python version, and other tools you might need
 build:
   os: ubuntu-24.04
-  tools:
-    python: "3.13"
```
nelimee, line 11: This is not needed because `pixi` brings its own version of `python`, right?

nelimee (commented):
At first glance, that looks good to me.

nelimee (approved):
I tested this PR locally and Arindam did that too on WSL, it worked in both cases. Because that is fixing a bug disrupting developer workflow, I am merging now.

PR #123, opened 2025-11-04
feat: adding QuOps computation methods. Task #122

### LINKED ISSUE #122: request: connect distance with logical error probability (LEP)
### Motivation

Add functions, which can, with a fixed `p` level, tell, how many QuOps can be achieved with a particular code distance; and vice versa: which code distance protects expected QuOps.

### Proposed solution

We have internal draft code. It will be adapted in `explorer.analysis` module.

### Potential alternatives

_No response_

### Additional context

_No response_

str-anger:
## 🔗 Closed Issues

"Closes #122".

---

## 📝 Description

This PR adds previously existing functions for computation of QuOps from distance and vice versa.

This PR uses a formula from the literature to compute QuOps for the distance.
And it solves an inequality min(D): `LEP(D | p0, lambda) < 1 / QuOps`. The function is not always monotonous, which means that binary search may find a sweet spot with D=1.

Implementations by @CaiReesManuelRL

---

## 🚦 Status

---

## 🛠️ Future Work

N/A

---

## ➕️ Additional Information

N/A

---

## 🧾 Release Note

This PR adds functions for computation of QuOps from rotated planar code distance; and vice versa.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "pip" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
--- a/deltakit-explorer/src/deltakit_explorer/analysis/__init__.py
+++ b/deltakit-explorer/src/deltakit_explorer/analysis/__init__.py
@@ -2,8 +2,14 @@
 """Description of ``deltakit.explorer.analysis`` namespace here."""
 
 from deltakit_explorer.analysis._analysis import (
-    calculate_lambda_and_lambda_stddev, calculate_lep_and_lep_stddev,
-    get_exp_fit, get_lambda_fit)
+    get_exp_fit, get_lambda_fit, calculate_lep_and_lep_stddev
+)
+
+from deltakit_explorer.analysis.lambda_ import calculate_lambda_and_lambda_stddev
+
+from deltakit_explorer.analysis._leppr import (
+    compute_logical_error_per_round, LogicalErrorProbabilityPerRoundResults,
+    simulate_different_round_numbers_for_lep_per_round_estimation)
 
 from deltakit_explorer.analysis._quops import \
     predict_distance_for_quops, predict_quops_at_distance
--- a/deltakit-explorer/src/deltakit_explorer/analysis/_analysis.py
+++ b/deltakit-explorer/src/deltakit_explorer/analysis/_analysis.py
@@ -1,18 +1,20 @@
 # (c) Copyright Riverlane 2020-2025.
-"""`analysis` module aggregates client-side analytical tools.
-"""
-from __future__ import annotations
-from collections.abc import Callable, Sequence
-from dataclasses import dataclass
-from math import floor
-import warnings
+"""`analysis` module aggregates client-side analytical tools."""
 
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+from deltakit_core.deprecation import deprecated
 import numpy as np
 import numpy.typing as npt
-from scipy.optimize import curve_fit
-from deltakit_explorer._utils._logging import Logging
+import semver
 
-
+@deprecated(
+    reason="A better function is now available.",
+    replaced_by="deltakit.explorer.analysis.compute_logical_error_per_round",
+    removed_in_version=semver.Version(1, 0, 0),
+)
 def get_exp_fit(
     logical_fails_all_rounds: npt.NDArray[np.int_] | list[int],
     shots_all_rounds: npt.NDArray[np.int_] | list[int],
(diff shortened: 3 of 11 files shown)


deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
```diff
@@ -0,0 +1,107 @@
...
+        Number of desired QuOps, must be a positive integer greater than 2.
+
+    Raises
+    ------
+    ValueError
+        - if solution is not found
+    """
+
+    if num_quops < 2:
+        raise ValueError("Number of QuOps should be at least 2")
```
nelimee: Maybe check that `lambda_ > 1` and if not output a warning/error saying that the target LEP cannot be reached?

nelimee (commented):
A few comments. I still need to convince myself that the bisection is correct (I already had issues with `//` rounding in a bisection, want to make sure that this implementation is correct).

deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
CaiReesManuelRL: I'm not sure if this is why I ommitted that check but arguably you could still have sufficiently low `p0` such that, even though `p0 * lambda0 ** (-(distance + 1) / 2)` is strictly decreasing with distance, you can achieve the desired QuOps. 

Maybe an optimistic QPU manufacturer wants to know how if they can do `x` QuOps at any distance :)

deltakit-explorer/tests/analysis/test_quops.py
```diff
@@ -0,0 +1,83 @@
...
+
+@pytest.mark.parametrize(
+    "p0, lambda_, distance, quops",
+    [
+        (1.0e-3, 2.0, 11, 5800),
+        # TODO: this test fails, as it actually finds
+        # a solution better: no QEC (d=1).
+        # in this case the function is not descending
+        # Question: how do I treat this?
+        # (1.0e-2, 1.5, 5, 65),
```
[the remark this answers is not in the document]
CaiReesManuelRL: For a bit of context:
The function plotted is the LEP of this dxdxd block. It turns out that there may be 0, 1 or 2 roots to the equation we are trying to solve, `LEP_of_dxdxd_block = target_LEP `. An analytical formula of this thing is not straightforward and involves some functions from the scipy library I wasn't comfortable using. Plus, we only need a coarse-grained estimate for distance, since in the end we can only build integer distance codes. 

Screenshot 2025-11-07 at 10 52 12

The above shows when the two root case can occur, when `LEP_of_dxdxd_block` is not monatonically decreasing, which happens for valid values of `lambda` and `lambda0`. There are two competing effects, increase code distance exponentially reduces `LEP_per_round`, but increasing  code distance also increases the number of rounds, more chances for error. 
Normally, the exponential suppression of the `LEP_per_round` always wins. But if `lambda` is sufficiently close to 1 and `lambda0` is above 0.5 (which I think is still a valid value from the fitting) then the increase in the number of rounds can win for a bit, LEP increases with d, and then goes back down once the exponential suppresion ramps up.

deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
```diff
@@ -0,0 +1,110 @@
...
+    distance = _equal_or_less_descending_bisection(
+        lambda x: _calculate_lep(p0, lambda0, x, x),
+        required_lep,
+        minimum=2,
+        maximum=999,
+    )
+    if distance is not None:
+        return distance
+    else:
+        raise ValueError("Could not find a solution for LEP(distance) < 1 / QoOps")
```
CaiReesManuelRL: In the above implementation does this only happen when the distance required exceeds 999? In which case it would be nice to tell a user that it could not find a solution in that range.

deltakit-explorer/pyproject.toml
```diff
@@ -41,7 +41,7 @@ dependencies = [
     "requests_toolbelt >=1.0.0",  # needed by gql, which doesn't automatically install it
     "pymatching >= 2.2.2",
     "galois >= 0.4.6",
-    "bposd >= 2.1",
+    "bposd @ git+https://github.com/quantumgizmos/bp_osd.git@8894ec654b24ae875c07e5a361dcae9a77d748ce",
```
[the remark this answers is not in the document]
nelimee: Nope, this is on `main` at the moment, and will disappear when this branch will be updated.

deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
```diff
@@ -0,0 +1,137 @@
...
+    lambda0 (float):
+        constant factor of lambda fit.
+    lambda_ (float):
+        Error suppression factor.
+    distance (int):
+        The distance at which to calculate the number of QuOps.
+    """
+    if distance % 2 == 0:
+        raise ValueError(
+            "This method gives correct estimation only at odd distances."
```
nelimee:
```suggestion
            "This method gives correct estimation only at odd distances. "
```

deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
```diff
@@ -0,0 +1,137 @@
...
+    lambda_ (float):
+        Error suppression factor.
+    num_quops (int):
+        Number of desired QuOps, must be a positive integer greater than 2.
+    max_distance (int):
+        maximum distance to consider. Default is 999.
+    Raises
+    ------
+    ValueError
+        - if solution is not found
```
nelimee: And also if pre-conditions on parameters not met, please add.

deltakit-explorer/tests/analysis/test_quops.py
```diff
@@ -0,0 +1,95 @@
...
+    predict_distance_for_quops, predict_quops_at_distance
+
+
+Parameters = namedtuple("Parameters", ["lambda0", "lambda_"])
+
+def alternative_lep_per_round(p_0: float, lambda_: float, d: int) -> float:
+    return p_0 * exp(-log(lambda_) * (d + 1) / 2)
+
+@pytest.fixture
+def default_parameters():
```
nelimee:
```suggestion
def default_parameters() -> Parameters:
```

deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+
+    if num_quops < 2:
+        raise ValueError("Number of QuOps should be at least 2")
+
+    if lambda_ <= 1.0:
+        raise ValueError("Lambda should be greater than 0 to ensure error suppression")
```
nelimee:
```suggestion
        raise ValueError("Lambda should be greater than 1 to ensure error suppression")
```

nelimee (changes requested):
A few changes, the rest seems fine to me.

deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
```diff
@@ -0,0 +1,142 @@
...
+    lambda0: float,
+    lambda_: float,
+    num_quops: float,
+    max_distance: int=999,
+) -> int:
+    """Returns the nearest odd distance that achieves the desired number of QuOps to one
+    decimal place. Uses the definition that the number of QuOps achievable at a
+    particular distance is 1 / pL, where pL is the probability of a logical error
+    occurring during a distance-D, D-round memory experiment
+    without state preparation or measurement error.
```
nelimee:
```suggestion
    """Returns the nearest odd distance that achieves the desired number of QuOps to one
    decimal place. 
    
    Uses the definition that the number of QuOps achievable at a
    particular distance is 1 / pL, where pL is the probability of a logical error
    occurring during a distance-D, D-round memory experiment
    without state preparation or measurement error.
```

deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Returns:
+        int | None:
+            a value, which safisfies the search, or None.
+    """
+    while minimum < maximum:
+        value = func(minimum)
+        if value <= target:
+            return minimum
+        minimum += 1
```
nelimee: Can be simplified (at least in my opinion that's a simplification) to:

```suggestion
    for test_input in range(minimum, maximum):
        value = func(test_input)
        if value <= target:
            return test_input
```

PR #126, opened 2025-11-05
chore: improve leppr function consistency

### LINKED ISSUE #125: request: improve logical error probability per round computation consistency
### Motivation

The function estimating logical error probability per round (LEPPR) in

https://github.com/Deltakit/deltakit/blob/e0223b5d5d28fd9d2b6a7f7596d343d16b76df77/deltakit-explorer/src/deltakit_explorer/analysis/_analysis.py#L157

could be made more consistent with changes that will be introduced by #68.

### Proposed solution

Here is a non-exhaustive list of changes:

- reduce the size of the large `_analysis.py` file by putting everything related to LEPPR to its own module (`_leppr.py`?).
- Remove the attributes `*_stddev_propagated` and `*_stddev_fit` from `LogicalErrorProbabilityPerRoundResults`.
- Do not compute logical error probability in the function but take it as an argument instead of the number of `shots` and `fails`.

### Potential alternatives

_No response_

### Additional context

_No response_

nelimee:
## 🔗 Closed Issues

Closes #125.

---

## 📝 Description

Try to address all the points from #125. 

A particular note on the use of `curve_fit`. Ideally, because we are doing polynomial fitting here, a call to `numpy.polyfit` seems like the best option. But in order to get the covariance matrix, `numpy.polyfit` requires 3 data points or more, which makes the function fail when only 2 data points are provided.
`scipy.optimize.curve_fit` was used successfully before, so keeping it for the moment.

---

## 🚦 Status

Ready for review (blocked by the same issue as #68).

---

## 🛠️ Future Work

None

---

## ➕️ Additional Information

This PR is composed of 2 commits:

1. moving the LEPPR-related functions to its own module (no new code here, no need to review that commit).
2. improving the consistency.

Even though the diff seems large, that is mostly due to 1) and there is no practical value in reviewing 1). So you can just review 2 and that should be enough.

---

## 🧾 Release Note

PR title

### DIFF
--- a/deltakit-explorer/pyproject.toml
+++ b/deltakit-explorer/pyproject.toml
@@ -41,7 +41,7 @@ dependencies = [
     "requests_toolbelt >=1.0.0",  # needed by gql, which doesn't automatically install it
     "pymatching >= 2.2.2",
     "galois >= 0.4.6",
-    "bposd >= 2.1",
+    "bposd @ git+https://github.com/quantumgizmos/bp_osd.git@8894ec654b24ae875c07e5a361dcae9a77d748ce",
     "pandas >= 2.0",
     "pandas-stubs",
     "types-seaborn",


AndrewPattersonRL (approved):
Thanks Adrien, I've reviewed, but no changes from me as you've documented the reasons here and in the linked issue well.

PR #130, opened 2025-11-06
chore: enforce more linting rules

nelimee:
**NOTE: do not review the full PR, read this first**

-------------

## 🔗 Closed Issues

Partial fix of #27

---

## 📝 Description

This PR adds a lot of linting rules to the repository and changes the code when it was not valid according to the added rules. 

Most of the changes have been done automatically, and you will see that most of them are trivial. 

I plan to do a full re-read of that PR before marking it ready for review, potentially flagging changes I am not sure about to simplify the review work. 
If you really want to review everything, doing so on a per-commit basis might be a good call, as each commit should be addressing similar issues.

---

## 🚦 Status

A few of the rules listed in #27 are not added yet. Will add them later.

---

## 🛠️ Future Work

None.

---

## ➕️ Additional Information

Rules in #27 that were not included:

- `ICN` because it raises a warning on `import matplotlib` and wants `import matplotlib as mpl`. I prefer to keep flexibility on such namings.
- `PT` because the amount of warnings is massive and it will require a lot of work to add all of them.
- `PTH` because it pushes for `Path.open` instead of `open` and that does not seem like a very valuable change for the moment.

---

## 🧾 Release Note

PR title

### DIFF
--- a/deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
+++ b/deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
@@ -7,7 +7,7 @@ References
 - https://doi.org/10.48550/arXiv.2408.13687
 """
 
-from typing import Callable
+from collections.abc import Callable
 
 
 def _equal_or_less_brute_force_search(
@@ -72,10 +72,11 @@ def predict_quops_at_distance(lambda0: float, lambda_: float, distance: int) -> 
         The distance at which to calculate the number of QuOps.
     """
     if distance % 2 == 0:
-        raise ValueError(
+        msg = (
             "This method gives correct estimation only at odd distances. "
             f"Distance provided: {distance}"
         )
+        raise ValueError(msg)
     return 1. / _calculate_lep(lambda0, lambda_, distance, distance)
 
 def predict_distance_for_quops(
@@ -112,10 +113,12 @@ def predict_distance_for_quops(
     """
 
     if num_quops < 2:
-        raise ValueError("Number of QuOps should be at least 2")
+        msg = "Number of QuOps should be at least 2"
+        raise ValueError(msg)
 
     if lambda_ <= 1.0:
-        raise ValueError("Lambda should be greater than 1 to ensure error suppression")
+        msg = "Lambda should be greater than 1 to ensure error suppression"
+        raise ValueError(msg)
 
     required_lep = 1. / num_quops
     distance = _equal_or_less_brute_force_search(
@@ -125,11 +128,10 @@ def predict_distance_for_quops(
         maximum=max_distance,
         step=2,  # seek for odd solutions
     )
-    if distance is not None:
-        return distance
-    else:
+    if distance is None:
         text = (
             f"Could not find a solution between [1, {max_distance}] "
             "for LEP(distance) < 1 / QuOps. Increase max_distance."
         )
         raise ValueError(text)
+    return distance
--- a/deltakit-explorer/tests/analysis/test_quops.py
+++ b/deltakit-explorer/tests/analysis/test_quops.py
@@ -1,15 +1,20 @@
 # (c) Copyright Riverlane 2020-2025.
 from __future__ import annotations
 
-from collections import namedtuple
 from math import exp, log
+from typing import NamedTuple
 
 import pytest
-from deltakit_explorer.analysis import \
-    predict_distance_for_quops, predict_quops_at_distance
+
+from deltakit_explorer.analysis import (
+    predict_distance_for_quops,
+    predict_quops_at_distance,
+)
 
 
-Parameters = namedtuple("Parameters", ["lambda0", "lambda_"])
+class Parameters(NamedTuple):
+    lambda0: float
+    lambda_: float
 
 def alternative_lep_per_round(p_0: float, lambda_: float, d: int) -> float:
     return p_0 * exp(-log(lambda_) * (d + 1) / 2)


deltakit-core/src/deltakit_core/decoding_graphs/_data_qubits.py
```diff
@@ -122,36 +114,35 @@ def __init__(self, vertices: Iterable[int]):
...
-        try:
-            return self._hash == __o._hash and self.vertices == __o.vertices
-        except AttributeError:
-            return False
+    def __eq__(self, other: object) -> bool:
+        return (
+            isinstance(other, DecodingHyperEdge)
+            and self._hash == other._hash
+            and self.vertices == other.vertices
+        )
```
nelimee: To double-check

deltakit-core/src/deltakit_core/decoding_graphs/_decoding_graph.py
```diff
@@ -153,35 +167,35 @@ class DecodingHyperMultiGraph(HyperMultiGraph[Tuple[DecodingHyperEdge, int]]):
...
                 edge = data
                 edge_record = EdgeRecord()
-            elif isinstance(data, tuple) and isinstance(data[-1], EdgeRecord):
-                edge, edge_record = data  # type: ignore
-            elif isinstance(data, tuple) and isinstance(data[-1], int):
-                edge = DecodingHyperEdge(data)  # type: ignore
+            elif _is_decoding_hyper_edge_and_edge_record(data):
+                edge, edge_record = data
+            elif _is_tuple_of_ints(data):
+                edge = DecodingHyperEdge(data)
```
nelimee: To double-check, introduction of `TypeGuard`, see https://typing.python.org/en/latest/spec/narrowing.html

tests/qmem/test_qmem.py
```diff
@@ -252,20 +247,13 @@ def plot_per_round_logical_error(data, colors, distance: int) -> None:
...
         x_distance=3, z_distance=3, num_rounds=3, physical_error_rate=0.001
-    ).run_batch_shots(1e5)
-    print(
-        f"Logical error probability with physical error 1%: {pct1_fails / pct1_shots}"
-    )
-    print(
-        "Logical error probability with physical error .1%:"
-        f"{pt1pct1_fails / pt1pct1_shots}"
-    )
+    ).run_batch_shots(10_000)
```
nelimee: Was the print really needed here? To check.

deltakit-explorer/tests/logging/test_logging.py
```diff
@@ -29,6 +30,7 @@ def setup_class(self):
         Logging.set_log_level(logging.INFO)
 
     def teardown_class(self):
+        self.logfile.close()
```
nelimee: This seems to have been forgotten. I added it just in case. To double-check.

deltakit-explorer/src/deltakit_explorer/codes/_css/_experiment_circuit.py
```diff
@@ -42,23 +43,27 @@ def experiment_circuit(experiment: Sequence[CSSStage]) -> Circuit:
...
+        raise ValueError(
+            msg
+        )
     circuit = experiment[0].remaining_rounds
 
     # Get the detectors for stage transitions.
     stage_transition_detectors = [
         get_stage_transition_circuit(previous_stage, current_stage)
-        for previous_stage, current_stage in zip(experiment[:-1], experiment[1:])
+        for previous_stage, current_stage in itertools.pairwise(experiment)
```
nelimee: Should be fine, to double-check.

deltakit-decode/tests/analysis/test_empirical_decoding_error_distribution.py
```diff
@@ -212,7 +211,6 @@ def test_batch_record_adds_correct_errors(self):
         distr = EmpiricalDecodingErrorDistribution(3)
 
         distr.batch_record_errors(corrections, target)
-        print(distr.to_dict())
```
nelimee: Double-check that this deletion is fine.

deltakit-circuit/src/deltakit_circuit/_circuit.py
```diff
@@ -200,9 +185,8 @@ def transform_qubits(self, id_mapping: Mapping[T, U]):
             A mapping of qubit types to other qubit types
         """
         if len(id_mapping.values()) != len(set(id_mapping.values())):
-            raise ValueError(
-                "The ID mapping is not bijective, all values must be unique."
-            )
+            msg = "The ID mapping is not bijective, all values must be unique."
+            raise ValueError(msg)
```
theinsanetramp: I'd love to know the justification for this linting rule

deltakit-decode/src/deltakit_decode/analysis/_decoder_manager.py
```diff
@@ -123,11 +121,11 @@ def clear_pool_manager(pool, total_processes: int):
...
-                                 batch_limit: Optional[int],
-                                 processes: int,
-                                 pool,
-                                 min_tasks_per_process: int = 50
-                                 ) -> Tuple[int, int]:
+                                 batch_limit: int | None,
+                                 processes: int,  # noqa: ARG002
+                                 pool,  # noqa: ARG002
+                                 min_tasks_per_process: int = 50  # noqa: ARG002
+                                 ) -> tuple[int, int]:
```
theinsanetramp: There's a `# pylint: disable=unused-argument` you've missed below this. Also these args smell - this is a non-abstract function in an ABC that has all these unused args, presumably so it can be overloaded?

deltakit-explorer/tests/unit/circuits/test_tableau_functions.py
```diff
@@ -5196,7 +5177,7 @@ def test_compile_circuit_with_tableau_compiles_snippets_of_memory_circuits_corre
             assert compiled_circ == expected_circ
 
         def test_no_non_native_gates_left_after_compilation(
-            self, circ, native_gate_set, expected_circ
+            self, circ, native_gate_set, expected_circ  # noqa: ARG002
         ):
```
theinsanetramp: Looks to me like `expected_circ` isn't used and should be removed - ARG002 may be right here, same at 5269

deltakit-circuit/src/deltakit_circuit/noise_channels/_abstract_noise_channels.py
```diff
@@ -43,7 +33,7 @@ class NoiseChannel(ABC, Generic[T]):
 
     stim_string: ClassVar[str]
 
-    def __init__(self, *args, tag: str | None = None, **kwargs) -> None:
+    def __init__(self, *_args, tag: str | None = None, **_kwargs) -> None:
         super().__init__()
         self._tag = tag
```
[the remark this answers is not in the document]
nelimee: They are used in child classes because of multi-inheritance shenanigans (call to `super().__init__` that needs to forward values...). Ideally, I would love to just remove these by removing any multi-inheritance. In practice, I don't feel like changing something that "works" and that may be replaced in the future.

Note that `__init__` here is not marked as abstract, so child classes can redefine it without making sure that the method signature exactly match.

deltakit-explorer/tests/unit/circuits/test_tableau_functions.py
nelimee: Yep, the problem is that the whole class is parameterized with `pytest.mark.parameterize` and some of the test methods in the class use that parameter. This method might be wrongly grouped, but I feel like this is too much of a change just for a "more lint rules" PR.

deltakit-explorer/src/deltakit_explorer/qpu/_circuits/_circuit_functions.py
```diff
@@ -143,7 +142,7 @@ def _calculate_offset_annotation_layers(
     # now remove one repeat from the repeat block if it is beneficial to do so
     new_merged_layers = []
     skip_layer = False
-    for layer, next_layer in zip(merged_layers[:-1], merged_layers[1:]):
+    for layer, next_layer in itertools.pairwise(merged_layers):
```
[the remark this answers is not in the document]
nelimee: c.f. my answer to my own comment in `_experiment_circuit.py`. This should be fine.

nelimee (commented):
Closing a few annotations after my own double-check. The ones left are the ones that would benefit from an external double-check.

deltakit-core/src/deltakit_core/decoding_graphs/_data_qubits.py
AndrewPattersonRL: I would presume it allows for `DecodingHyperEdge` to be the same structure but not the same class - it may be that we are allowing comparison to a hyper edge generated in some other way, perhaps `stim`? However I didn't write this, and as it is requires us to keep up to date with some (unknown) external package. I think this is fine, making the conversion to `DecodingHyperEdge` explicit.

nelimee:
Changes with merge were trivial, 2 approvals, I am merging this one. Thanks for the reviews!

PR #149, opened 2025-12-16
chore: use uv and remove pixi

nelimee:
## 🔗 Closed Issues

This should eventually close #145 

---

## 📝 Description

This PR removes all uses of `pixi` and replaces it by `uv`.

It also simplifies the CI by removing some workflows.

---

## 🚦 Status

- CI probably needs more care. Extensive test should be performed to check that we are not loosing anything in the move.
- Package versions should be extensively reworked: now that we are correctly managing dependencies, a lot of mis-configurations are appearing and need to be fixed.

---

## 🛠️ Future Work

None

---

## ➕️ Additional Information

None

---

## 🧾 Release Note

PR title

Issue #51, opened 2025-09-21
request: improve module import times


mdhaber:
### Motivation

Some modules take a long time to import. For instance:

<img width="877" height="675" alt="Image" src="https://github.com/user-attachments/assets/0c9fe4f8-e1a0-4bca-b79d-154140f704af" />

### Proposed solution

Adopt [SPEC-0001](https://scientific-python.org/specs/spec-0001/); use [`lazy-loader`](https://github.com/scientific-python/lazy-loader).

rolandriver:
This must be prioritised as all subpackages imports from the metapackage take unusually long. Tagging @nelimee for reference.

PR #151, opened 2025-12-19
fix: tag feature when exporting with stim<1.15

nelimee:
## 🔗 Closed Issues

None, that's a small fix for an issue found in https://github.com/Deltakit/deltakit/pull/149

---

## 📝 Description

The following python code was failing if a version of `stim` without the tag feature was installed:

```py
from deltakit_circuit import Circuit, GateLayer
from deltakit_circuit.gates import X

circuit: Circuit[int] = Circuit()
circuit.append_layers([GateLayer(X(0, tag="hello"))])
print(circuit.as_stim_circuit())
```

This PR fixes that by conditionally including the tag if `stim` is known to support that (according to the package version).

---

## 🚦 Status

Ready for review.

---

## 🛠️ Future Work

None

---

## ➕️ Additional Information

None

---

## 🧾 Release Note

PR title

### DIFF
--- a/deltakit-circuit/tests/test_stim_version_compatibility.py
+++ b/deltakit-circuit/tests/test_stim_version_compatibility.py
@@ -1,12 +1,12 @@
-import semver
 import stim
+from packaging.version import Version
 
 from deltakit_circuit._stim_version_compatibility import is_stim_tag_feature_available
 
 
 def test_stim_version_compatibility() -> None:
-    current_stim_version = semver.Version.parse(stim.__version__)
-    if current_stim_version >= semver.Version(1, 15):
+    current_stim_version = Version(stim.__version__)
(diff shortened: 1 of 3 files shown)


deltakit-circuit/tests/test_stim_version_compatibility.py
```diff
@@ -0,0 +1,12 @@
+import semver
+import stim
```
rolandriver: Is it worth importing `stim` here or using the `import.metadata` trick above will do ?

nelimee: Even though both should be equivalent, I prefer to use a different method just to make sure that all methods agree.

rolandriver: I guess this will be resolved at a later point in #51.

PR #154, opened 2025-12-31
ci: update release.yml with broader tests

### LINKED ISSUE #19: request: test wheels on additional OS/architecture combinations
### Motivation

Now that the repo is public, we have additional OS/architecture options to use in workflows. See [Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories).

### Proposed solution

Update `release.yml` to include these combinations.

### Additional context

@chididi IIRC, we excluded combinations from the matrix only because they were not available on private repos?

https://github.com/Deltakit/deltakit/blob/717fc7f4e7110d95e25f5793838b0d94f5f9b2a9/.github/workflows/release.yml#L60-L73

Any reason not to use other combinations that are available now?

### LINKED ISSUE #39: request: run full test suite against wheels in release workflow(s)
### Motivation

The release workflow only checks that wheels import with different OS/architecture combinations.

### Proposed solution

Consider running the full test suite, too.

### Additional context

This might be a good place to address gh-37 (but that could also be done in PR CI)

nelimee:
## 🔗 Closed Issues

Fixes #19 by considering more runners in CI.
Fixes #39 by testing the wheels on all the runners.

---

## 📝 Description

This PR rewrites the `release.yml` workflow to:

1. use `uv` to make the workflow file simpler and harmonise workflows,
2. test all `deltakit` packages on more runners than our usual `pull_request` workflow does, to double-check that we did not miss something before releasing,
3. use more runners than before, including ubuntu arm and windows arm.

My opinion is that we do not have to include `macos-15-intel` runner, but my arguments are weak so I am open to include it if there are valid concerns.

---

## 🚦 Status

Ready for review.

---

## 🛠️ Future Work

None.

---

## ➕️ Additional Information

None

---

## 🧾 Release Note

PR title


.github/workflows/release.yml
```diff
@@ -23,120 +23,104 @@ on:
...
-        project: [deltakit, deltakit-circuit, deltakit-core, deltakit-decode, deltakit-explorer]
-    steps:
-      - uses: actions/checkout@v4
 
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-          python-version: '3.10'
+  # Build all deltakit wheels and sdists.
+  build_dists:
```
rolandriver:
```suggestion
  build-wheels-sdists:
```

.github/workflows/release.yml
```diff
@@ -23,120 +23,104 @@ on:
...
-  test-wheel:
+          name: all-deltakit-dists
+          path: ./dist/*
+          # No need to keep these as they will be published in this workflow.
+          retention-days: 1
+
+  # In order to avoid creating too much CI jobs, and because most packages will be
+  # installed through a "pip install deltakit", this job does not test each deltakit
+  # package in isolation but rather installs all of them and performs all the tests in
+  # the same environment.
```
rolandriver:
```suggestion
  # To avoid creating unnecessary CI jobs, and because most packages are 
  # typically installed through `pip install deltakit`, this job does not test each deltakit
  # package in isolation. Instead, it installs all of them and performs all tests suites in
  # a single environment.
```

.github/workflows/release.yml
```diff
@@ -23,120 +23,104 @@ on:
...
+          #   arch: x64
+        python-version: ['3.10', '3.11', '3.12', '3.13']
+        dependency-resolution: ['highest', 'lowest-direct']
     steps:
-      - name: Set up Python
-        uses: actions/setup-python@v5
-        with:
-            python-version: ${{ matrix.python-version }}
-            architecture: ${{ matrix.arch }}
+      - uses: actions/checkout@v6
```
rolandriver: Name.

.github/workflows/release.yml
```diff
@@ -23,120 +23,104 @@ on:
...
+          - os:	ubuntu-24.04-arm
             arch: arm64
           - os: windows-latest
+            arch: x64
+          - os: windows-11-arm
             arch: arm64
           - os: macos-latest
-            arch: x64
+            arch: arm64
+          # Not including Intel-based MacOS, should it be?
```
rolandriver: If we do, we should also include this platform in tests.

.github/workflows/release.yml
```diff
@@ -23,120 +23,104 @@ on:
...
       matrix:
-        arch: [x64, arm64]
-        python-version: ['3.10', '3.11', '3.12', '3.13']
-        os:
-          - ubuntu-latest
-          - macos-latest
-          - windows-latest
-        exclude:
+        os_arch:
           - os: ubuntu-latest
```
rolandriver: `ubuntu-slim` ?

rolandriver (commented):
Thanks @nelimee. Comments form my side.

.github/workflows/release.yml
nelimee: This builds both the source distribution and the wheels, hence why the name is more generic.

.github/workflows/release.yml
nelimee: We could, but `ubuntu-slim` is supposed to be for lightweight tasks, and this is launching all the tests in the test suite, for all packages, which is not really lightweight. I think the `ubuntu-latest` is better suited.

.github/workflows/release.yml
```diff
@@ -23,120 +23,104 @@ on:
...
+          name: all-deltakit-dists
+          path: ./dist/*
+          # No need to keep these as they will be published in this workflow.
+          retention-days: 1
+
+  # In order to avoid creating too much CI jobs, and because most packages will be
+  # installed through a "pip install deltakit", this job does not test each deltakit
+  # package in isolation but rather installs all of them and performs all the tests in
+  # the same environment.
+  test_wheels:
```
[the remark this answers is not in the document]
nelimee: None of our current workflow use hyphens at the job level. I think this change should be done in a separate PR if it is really needed.

.github/workflows/release.yml
```diff
@@ -23,120 +23,104 @@ on:
...
-        working-directory: ./project
-        run: |
-          python -m venv .venv
-          source .venv/bin/activate
-          pip install *.tar.gz
-          python -c "import deltakit, deltakit_circuit, deltakit_core, deltakit_decode, deltakit_explorer"
+      - name: Try to install the sdists
+        run: uv pip install ./dist/deltakit*.tar.gz
 
   publish:
```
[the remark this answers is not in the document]
nelimee: I think that the version tag on PyPI is deduced from the metadata in the uploaded files, so this has nothing to do with version as far as I am aware.
But there is a need to only publish a few of the packages, for example if we only bumped `deltakit` and `deltakit-circuit` versions. Any idea for that?

.github/workflows/release.yml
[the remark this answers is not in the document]
rolandriver: Actually, there is `github-release` further down this file. So yeah I would advise to harmonise.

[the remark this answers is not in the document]
rolandriver: Sure OK for another PR.

PR #164, opened 2026-01-08
ci: simplifying CI by grouping tests

### LINKED ISSUE #159: bug: actions in CI are not correctly filtered out
### Bug Description

Static checks and tests are supposed to only be performed on packages that are changed by the PR. This appears to not be true.

### Minimal Reproducible Example (MRE)

Taking https://github.com/Deltakit/deltakit/pull/157 as an example:

- labelling seems fine: only `deltakit-circuit` was changed.
- but the jobs executed in CI tested all `deltakit` packages, as can be seen in [this run](https://github.com/Deltakit/deltakit/pull/157/checks?sha=182c452f3d826a1ffa58dd6e360b80d14134158b) for example.

### Actual behaviour

_No response_

### Expected behaviour

Only `deltakit-circuit`-related CI jobs should have been fired by the CI in https://github.com/Deltakit/deltakit/pull/157/checks?sha=182c452f3d826a1ffa58dd6e360b80d14134158b.

### Additional context

_No response_

nelimee:
## 🔗 Closed Issues

Closes #159 by making it not relevant any more if merged. 

This is an exploratory PR that needs to be pushed and opened because it explores new ways to organise the CI with GitHub actions.

---

## 📝 Description

This PR:

- removes most of the complexity regarding tagging packages (they are still tagged, but CI does not try to recover the tags to filter out steps),
- reduces the amount of CI jobs that are spawned at each commit.

---

## 🚦 Status

Ready for review.

---

## 🛠️ Future Work

Introduce coverage reports in CI.

---

## ➕️ Additional Information

None

---

## 🧾 Release Note

None

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,7 +124,13 @@ testpaths = [
 ]
 addopts = ["-ra", "--showlocals", "--strict-markers", "--strict-config"]
 xfail_strict = true
-filterwarnings = ["error"]
+filterwarnings = [
+  "error",
+  # Note: the below line ignore a specific warning about changes in multiprocessing.
+  # This warning is only raised by a few tests in deltakit-decode. This is a temporary
+  # fix.
+  "ignore:This process .* is multi-threaded, use of fork\\(\\) may lead to deadlocks in the child.:DeprecationWarning",
+]
 
 [tool.bandit]
 # './tools/vale.py' is for `pixi run bandit`; 'tools/vale.py' is for pre-commit


.github/workflows/pull_request.yml
```diff
@@ -80,25 +43,26 @@ jobs:
       - name: Download wheels
         uses: actions/download-artifact@v4
         with:
-          name: all-deltakit-wheels
+          name: deltakit-wheels
           path: ./dist
-      - name: Setup and sync venv for testing ${{ matrix.projects.project }}
```
rolandriver: I still find it useful for the same reason stated above.

PR #271, opened 2026-06-04
feat: visualise detection probability on patch (#135)

### LINKED ISSUE #135: request: visualise detection probability on patch
### Motivation

`deltakit` is able to plot detection probability (also called defect rate) as:

Image

(from https://deltakit-docs.riverlane.com/en/docs/examples/notebooks/google_qmem_experiment/noise_analysis.html)

It would be nice to also be able to plot that on a code patch when it makes sense. For example, when the rotated surface code is used, detectors can be "assigned" to plaquettes which leads to plots like

Image

(image from https://www.nature.com/articles/s41586-022-05434-1/figures/2)

### Proposed solution

Implement the above visualisation at least for the rotated surface code. We might want to be able to plot:

- for each round,
- an average of all the rounds,
- a median of all the rounds,
- variance of all the rounds except the first and last ones (which are expected to be outliers),
- [more ideas?]

### Potential alternatives

_No response_

### Additional context

_No response_

Shanwis:
### 🔗 Closed Issues

Closes #135

### 📝 Description

Adds `plot_detection_probability_on_patch()` to overlay detection probabilities as coloured circles on a rotated surface code patch. Each stabiliser ancilla qubit is drawn with a colour mapped from the aggregated defect rate at that location.

Supports three aggregation modes:
- `"mean"` — mean across all rounds
- `"median"` — median across all rounds  
- `"variance"` — variance excluding first/last outlier rounds

The function accepts optional `fig`/`ax` for embedding as an inset plot, and `vmin`/`vmax`/`cmap`/`show_colorbar` for full visual control.

This contribution is part of [unitaryHACK 2026](https://unitaryhack.dev/projects/deltakit/).

**New files:**
- `deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py`
- `deltakit-explorer/tests/plotting/test_detection_on_patch.py`

**Modified files:**
- `deltakit-explorer/src/deltakit_explorer/plotting/__init__.py` (export)

### 🚦 Status

Ready for review.

### 🛠️ Future Work

- Support per-round subplot mode
- Extend to unrotated and toric code variants
- Add optional stabiliser-plaquette fill overlay (vs. current ancilla-circles)

### 🧾 Release Note

feat: visualise detection probability on patch (#135)


deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,187 @@
...
+    for coord, rates in detection_probabilities.items():
+        values = np.asarray(rates)
+        match mode:
+            case "average":
+                result[coord[:2]] = float(np.mean(values))
+            case "median":
+                result[coord[:2]] = float(np.median(values))
+            case "variance":
+                if len(values) > 2:
+                    values = values[1:-1]
```
nelimee: Why do you need that here and not for the other ones?

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,187 @@
...
+        code: The planar code whose stabiliser ancilla coordinates are used
+            for matching.
+        tolerance: Absolute tolerance for matching floating-point coordinates.
+
+    Returns:
+        Mapping from ancilla (x, y) coordinate pairs to their matched probability.
+    """
+    prob_by_xy: dict[tuple[float, float], float] = {}
+    for prob_coord, prob_val in aggregated.items():
+        prob_by_xy[(round(prob_coord[0], 6), round(prob_coord[1], 6))] = prob_val
```
nelimee: I am not sure if the `round` calls have any use here. We are not expecting multiple coordinates to clash, so that's not an "aggregation" mechanism, and you are computing absolute tolerances between two coordinates to check if they are considered equal. Why the round then?

nelimee (commented):
This seems OK to me. A few overall comments too:

- Can you include an example of a plot obtained by your code in the PR description (and the code to generate it if possible)?
- Initially I had in mind a colour that would fill the stabilisers. But that might be harder to do than simply plotting a point per syndrome qubit. What do you think?

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
Shanwis: It was mentioned in the issue spec 'variance of all the rounds except the first and last ones (which are expected to be outliers)'

Shanwis:
1. detection_on_patch

Code used:
```python3
import matplotlib.pyplot as plt
import numpy as np
from deltakit_explorer.codes import RotatedPlanarCode
from deltakit_explorer.plotting import plot_detection_probability_on_patch

code = RotatedPlanarCode(3, 3)
rng = np.random.default_rng(42)
ancilla_coords = [(0, 2), (2, 2), (2, 4), (2, 6), (4, 0), (4, 2), (4, 4), (6, 4)]
rates = {c: list(rng.uniform(0.02, 0.25, size=8)) for c in ancilla_coords}

fig, ax = plot_detection_probability_on_patch(code, rates, mode="average")
fig.savefig("detection_on_patch.png", bbox_inches="tight")
```

The 8 ancilla positions correspond to the 4 X-type and 4 Z-type stabilisers of the d=3 rotated surface code. Each circle is coloured by the mean detection probability across 8 rounds of synthetic data. The colourbar and patch legend are included automatically.

2. I went with ancilla circles since it was quicker to implement and avoids modifying _draw_code() — that way existing patch drawings stay untouched. I understand the initial spec envisioned coloured stabiliser plaquettes; I'm happy to explore that as a follow-up :). What do you think?

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,187 @@
...
+
+from deltakit_explorer.plotting._draw import _draw_code
+
+if TYPE_CHECKING:
+    from deltakit_explorer.codes._planar_code import PlanarCode
+
+
+def _aggregate_probabilities(
+    detection_probabilities: dict[tuple[float, ...], list[float]],
+    mode: Literal["average", "median", "variance"],
```
rolandriver: I generally prefer using enums to avoid syntactic errors.

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,187 @@
...
+            - ``"median"``: median of all rounds.
+            - ``"variance"``: variance excluding first and last rounds.
+
+    Returns:
+        Mapping from (x, y) coordinate pairs to the aggregated probability value.
+    """
+    result: dict[tuple[float, float], float] = {}
+    for coord, rates in detection_probabilities.items():
+        values = np.asarray(rates)
+        match mode:
```
rolandriver, line 121: There should be a fall-through case that raises an exception.

rolandriver (commented):
Thank you @Shanwis for you interest in contributing to Deltakit, this is much appreciated. I think this can be improved by overlaying the overall plaquette as in the example shown in the issue. I find circles in the middle of the plaquette a bit hard to read. If you want to respect stabiliser types, I'd suggest using two color palettes: for instance red shading for X-stabilisers and blue shading for Z-stabilisers. What do you think @nelimee ?

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
nelimee: That's correct, my bad I forgot about it! The first and last round are indeed outliers. The more I think about it, the more I think we should just provide a flag `include_outlier_rounds` (or any other name that makes sense) for the user to be able to pick whether or not to include these rounds, independently of the metric picked (average, median or variance). I can't remember why I specifically mentioned these outliers for the variance and not for the other ones. Give me a day to think this through, I'll keep that comment unresolved, and update you here.

nelimee:
> I went with ancilla circles since it was quicker to implement and avoids modifying _draw_code() — that way existing patch drawings stay untouched. I understand the initial spec envisioned coloured stabiliser plaquettes; I'm happy to explore that as a follow-up :). What do you think?

That makes sense to me, I think that was a good choice to start simple and see if that was enough.

I agree with @rolandriver: to me, the circles are not easy to parse, and will likely end up being too small when that plot will be included as an inset to another plot. So I think we should try to be able to colour plaquettes directly.

Now about the colour scheme, I don't know if splitting between shades or red and blue will be useful here. I am worried that it might blur some obvious spatial correlations just because the colours are not fit. I am not sure though, so my opinion is:

- if you want to explore the idea, probably try to play on the luminance of the colours to make sure that any given probability on patches of different types will end up with the same luminance.
- it will probably be simpler to not care about stabiliser types and only use a single colour map.

Shanwis:
I will try it out and push

deltakit-explorer/tests/plotting/test_detection_on_patch.py
```diff
@@ -0,0 +1,116 @@
...
+    (4.0, 0.0),
+    (4.0, 2.0),
+    (4.0, 4.0),
+    (6.0, 4.0),
+]
+NUM_ROUNDS = 8
+
+
+@pytest.fixture
+def code():
```
rolandriver: Return type annotations.

deltakit-explorer/tests/plotting/test_detection_on_patch.py
```diff
@@ -0,0 +1,116 @@
...
+    return RotatedPlanarCode(3, 3)
+
+
+@pytest.fixture
+def detection_probabilities():
```
rolandriver: Same.

deltakit-explorer/tests/plotting/test_detection_on_patch.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def test_custom_fig_ax(self, code, detection_probabilities):
+        fig, ax = plt.subplots()
+        fig_out, ax_out = plot_detection_probability_on_patch(
+            code, detection_probabilities, fig=fig, ax=ax
+        )
+        assert fig_out is fig
+        assert ax_out is ax
+
+    def test_average_mode(self, code, detection_probabilities):
```
rolandriver: Shouldn't you test the aggregation function itself instead of testing it through `plot_detection_probability_on_patch` ?

rolandriver (commented):
Thanks @Shanwis. Are these changes effectively use the overlay for the plaquette or is it still for circle in the middle of it? Is it too difficult to overlay the plaquette itself?

Shanwis:
Hi @rolandriver , sorry for the delay, I have made it so that the plot is similar to the example given in the original issue specs. 

The updated plot now:
- renders weight-4/interior stabilisers as square heatmap cells,
- renders weight-2/boundary stabilisers as edge-attached circles,
- removes data-qubit markers, ancilla markers, X/Z stabiliser background colours, and the patch legend,
- uses a single colormap to preserve spatial correlations,
- places the colourbar horizontally below the heatmap.

This should make the plot much easier to read as a small inset and closer to the style shown in the issue / Google paper figure.

Example code:

```python
import numpy as np
from deltakit_explorer.codes import RotatedPlanarCode
from deltakit_explorer.plotting import plot_detection_probability_on_patch

code = RotatedPlanarCode(5, 5)
rng = np.random.default_rng(42)

rates = {}
for stabilisers in code._stabilisers:
    for stabiliser in stabilisers:
        if stabiliser.ancilla_qubit is None:
            continue
        coord = stabiliser.ancilla_qubit.unique_identifier
        rates[(float(coord.x), float(coord.y))] = list(
            rng.uniform(0.10, 0.25, size=8)
        )

fig, ax = plot_detection_probability_on_patch(
    code,
    rates,
    mode="average",
    vmin=0.10,
    vmax=0.25,
)
fig.savefig("detection_on_patch_heatmap.png", bbox_inches="tight", dpi=200)
```
detection_on_patch_heatmap

Shanwis:
I also added direct unit tests for `_aggregate_probabilities` (6 tests covering average, median, variance with/without outlier removal, multiple coords, and unknown mode) as suggested in the review, and additionally for `_match_ancilla_coords` (4 tests covering basic matching, no-match, partial match, and tolerance), I had missed it earlier.

Shanwis:
I've updated the implementation with two visual styles:
* style="plaquette" (new default) — fills the actual stabiliser plaquettes (diamond-shaped for weight-4 interior, triangular for weight-2 boundary), matching the code geometry.
* style="heatmap" (previous approach) — simplified grid of square cells + edge-attached circles, useful for clean small insets.

style_comparison
Comparison (d=5, RotatedPlanarCode, random data)

Code:
```python3
import numpy as np
import matplotlib.pyplot as plt
from deltakit_explorer.codes import RotatedPlanarCode
from deltakit_explorer.plotting import plot_detection_probability_on_patch

code = RotatedPlanarCode(5, 5)
rng = np.random.default_rng(42)

rates = {}
for stabilisers in code.stabilisers:
    for stabiliser in stabilisers:
        anc = stabiliser.ancilla_qubit
        if anc is None:
            continue
        coord = anc.unique_identifier
        rates[(float(coord.x), float(coord.y))] = list(
            rng.uniform(0.10, 0.25, size=8)
        )

# Default plaquette style (single call, standalone)
fig, ax = plot_detection_probability_on_patch(code, rates, mode="average")
fig.savefig("detection_plaquette.png", bbox_inches="tight", dpi=150)

# Side-by-side comparison
fig, (ax1, ax2, ax3) = plt.subplots(1, 3, figsize=(15, 4))

plot_detection_probability_on_patch(
    code, rates, fig=fig, ax=ax1, style="plaquette",
    show_data_qubits=True, vmin=0.10, vmax=0.25, show_colorbar=False,
)
ax1.set_title('style="plaquette", show_data_qubits=True')

plot_detection_probability_on_patch(
    code, rates, fig=fig, ax=ax2, style="plaquette",
    vmin=0.10, vmax=0.25, show_colorbar=False,
)
ax2.set_title('style="plaquette" (default)')

plot_detection_probability_on_patch(
    code, rates, fig=fig, ax=ax3, style="heatmap",
    vmin=0.10, vmax=0.25, show_colorbar=True,
)
ax3.set_title('style="heatmap"')

fig.savefig("style_comparison.png", bbox_inches="tight", dpi=150)
```

The default is now style="plaquette", which is what the original issue spec envisioned. Data qubit markers are hidden by default but available via show_data_qubits=True.

One remaining open thread: @nelimee, any thoughts on the include_outlier_rounds flag?

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,368 @@
...
+                raise ValueError(msg)
+    return result
+
+
+def _match_ancilla_coords(
+    aggregated: dict[tuple[float, float], float],
+    code: PlanarCode,
+    tolerance: float = 1e-3,
+) -> dict[tuple[float, float], float]:
+    """Match aggregated probabilities to stabiliser ancilla coordinates.
```
rolandriver:
```suggestion
    """Match aggregated detection probabilities to stabiliser ancilla coordinates.
```

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,368 @@
...
+    Args:
+        aggregated: Mapping from (x, y) coordinate pairs to probability values.
+        code: The planar code whose stabiliser ancilla coordinates are used
+            for matching.
+        tolerance: Absolute tolerance for matching floating-point coordinates.
+
+    Returns:
+        Mapping from ancilla (x, y) coordinate pairs to their matched probability.
+    """
+    ancilla_values: dict[tuple[float, float], float] = {}
```
rolandriver:
```suggestion
    ancilla_det_prob: dict[tuple[float, float], float] = {}
```

Or smtg around this idea.

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,368 @@
...
+        anc = stabiliser.ancilla_qubit.unique_identifier
+        coords.append((float(anc.x), float(anc.y)))
+    if len(coords) < 3:
+        msg = "A stabiliser plaquette needs at least two data qubits."
+        raise ValueError(msg)
+
+    center = np.mean(coords, axis=0)
+    return sorted(
+        coords,
+        key=lambda p: np.arctan2(p[1] - center[1], p[0] - center[0]),
```
rolandriver, line 205: What is this sorting key doing?

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,368 @@
...
+    """Map a stabiliser to a square-grid heatmap position."""
+    if stabiliser.ancilla_qubit is None:
+        msg = "Cannot draw a detector heatmap for stabilisers without ancilla qubits."
+        raise ValueError(msg)
+
+    anc = stabiliser.ancilla_qubit.unique_identifier
+    weight = len([pauli for pauli in stabiliser.paulis if pauli is not None])
+
+    if weight == 4:
+        return "square", float(anc.x / 2 - 1), float(anc.y / 2 - 1)
```
rolandriver: I would create a custom datatype for this as it is a bit mysterious what these return types are intended to do.

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,368 @@
...
+    _vmax = vmax if vmax is not None else float(np.max(values))
+
+    if _vmax - _vmin < 1e-12:
+        _vmin = _vmin - 0.5
+        _vmax = _vmax + 0.5
+
+    norm = plt.Normalize(vmin=_vmin, vmax=_vmax)
+    cmap_obj = plt.get_cmap(cmap)
+
+    if style == "plaquette":
```
rolandriver, line 525: This function is quite long. I'd suggest to break it into specialiser plotter ones.

Shanwis:
Hello, I have fixed the workflow issues except the build-docs failing, as it is being addresses in PR #280

rolandriver (commented):
This is looking nice. Thank you. Some more comments from me.

Shanwis:
@rolandriver I believe I have made the changes mentioned in the comments

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,446 @@
...
+
+    patch_type: str
+    x: float
+    y: float
+
+
+def _heatmap_patch_position(
+    stabiliser: Stabiliser,
+    code: PlanarCode,
+) -> HeatmapPosition:
```
nelimee, line 226: Nitpick

```suggestion
def _heatmap_patch_position(stabiliser: Stabiliser, code: PlanarCode) -> HeatmapPosition:
```

I think this works in the line length limit

deltakit-explorer/src/deltakit_explorer/plotting/_detection_on_patch.py
```diff
@@ -0,0 +1,446 @@
...
+    """Position and shape of a stabiliser in the heatmap grid.
+
+    Attributes:
+        patch_type: ``"square"`` for interior weight-4 or ``"circle"`` for
+            boundary weight-2.
+        x: Heatmap column index.
+        y: Heatmap row index.
+    """
+
+    patch_type: str
```
nelimee:
```suggestion
    patch_type: Literal["square", "circle"]
```

nelimee (dismissed):
I really like that PR. For my comment about outliers, I think that having a flag controlling whether outliers are removed or not (independently of the aggregation function) is the way to go.

Provided that the above is done, and because my comments are very small, I'll pre-approve that PR. I am happy to merge once they are addressed if @rolandriver is also happy.

Well done!

Shanwis:
Switched mode from Literal["average", "median", "variance"] to a DetectionProbabilityAggregation enum, had forgot about it earlier :sweat: :

```python3
class DetectionProbabilityAggregation(str, Enum):
    MEAN = "mean"
    MEDIAN = "median"
    VARIANCE = "variance"
```

Using the str, Enum mixing

Also renamed to MEAN instead of AVERAGE, since mean is more commonly used (eg:- np.mean() in the codebase). 

@rolandriver

Shanwis:
Hey, I have just now added a few more stuff, 

- round_index: int | None — plots a single specific round instead of aggregating across all rounds. Supports negative indexing (round_index=-1 for last round). Takes precedence over mode when set.
- Input validation - a dedicated _validate_detection_probabilities() function checks all inputs upfront and reports every issue at once (non-finite values, out-of-range, wrong dimensions, short coordinates, etc.), instead of failing one-by-one. 

## Comparison figures

Code:

```python3
"""Generate comparison figures for all plot styles and modes."""

import numpy as np
import matplotlib.pyplot as plt
from deltakit_explorer.codes import RotatedPlanarCode
from deltakit_explorer.plotting import (
    DetectionProbabilityAggregation,
    plot_detection_probability_on_patch,
)

code = RotatedPlanarCode(5, 5)
rng = np.random.default_rng(42)

rates = {}
for stabilisers in code.stabilisers:
    for stabiliser in stabilisers:
        anc = stabiliser.ancilla_qubit
        if anc is None:
            continue
        coord = anc.unique_identifier
        rates[(float(coord.x), float(coord.y))] = list(
            rng.uniform(0.05, 0.30, size=10)
        )

# === Figure 1: 3×2 grid (modes × styles) ===
fig, axes = plt.subplots(2, 3, figsize=(14, 8))
modes = [
    DetectionProbabilityAggregation.MEAN,
    DetectionProbabilityAggregation.MEDIAN,
    DetectionProbabilityAggregation.VARIANCE,
]

for col, mode in enumerate(modes):
    # Top row: plaquette
    plot_detection_probability_on_patch(
        code, rates, mode=mode, style="plaquette",
        ax=axes[0, col], show_colorbar=False,
    )
    axes[0, col].set_title(f"Plaquette, {mode.value}", fontsize=11)

    # Bottom row: heatmap
    plot_detection_probability_on_patch(
        code, rates, mode=mode, style="heatmap",
        ax=axes[1, col], show_colorbar=False,
    )
    axes[1, col].set_title(f"Heatmap, {mode.value}", fontsize=11)

fig.savefig("comparison_modes_styles.png", bbox_inches="tight", dpi=150)
print("Saved comparison_modes_styles.png")

# === Figure 2: round_index example ===
fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(10, 4))

plot_detection_probability_on_patch(
    code, rates, round_index=0, style="plaquette",
    ax=ax1, show_colorbar=True,
)
ax1.set_title("Round 0 (plaquette)", fontsize=11)

plot_detection_probability_on_patch(
    code, rates, round_index=-1, style="heatmap",
    ax=ax2, show_colorbar=True,
)
ax2.set_title("Last round (heatmap)", fontsize=11)

fig.savefig("comparison_round_index.png", bbox_inches="tight", dpi=150)
print("Saved comparison_round_index.png")

# === Figure 3: with/without data qubits ===
fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(10, 5))

plot_detection_probability_on_patch(
    code, rates, style="plaquette",
    show_data_qubits=True, ax=ax1, show_colorbar=False,
)
ax1.set_title("With data qubits", fontsize=11)

plot_detection_probability_on_patch(
    code, rates, style="plaquette",
    show_data_qubits=False, ax=ax2, show_colorbar=False,
)
ax2.set_title("Without data qubits", fontsize=11)

fig.savefig("comparison_data_qubits.png", bbox_inches="tight", dpi=150)
print("Saved comparison_data_qubits.png")

plt.close("all")
print("Done")
```
comparison_data_qubits
comparison_modes_styles
comparison_round_index


With this I believe the feature is implemented :smile:

Shanwis:
@rolandriver Any other changes required?

The check-documentation-links workflow fail is due to a timeout and will resolve itself when run once more. The other is being fixed in PR #288 I believe.

@nelimee had approved the changes earlier.

rolandriver (approved):
Thanks @Shanwis this is looking good.

Shanwis:
The check-documentation-links failed due to the set timeout of 60s, it will work correctly if ran one more time.
Other than that all the checks pass.
All other checks are passing, and the same workflow completed successfully in my test PR where I validated the changes before submission.
I have added an issue to increase the timeout duration in issue #290.
image

Thank you very much for all your reviews and support @rolandriver @nelimee :heart:

Shanwis:
I have created PR #291 to address the intermittent timeout issue. Rerunning the workflow usually resolves the failure. Other than that, all checks are passing, and I believe this PR is ready for merge.

rolandriver:
Hi @Shanwis I approved and merged #291. Could you please rebase your branch against `main` and finalise this PR?

Shanwis:
Done @rolandriver

PR #273, opened 2026-06-04
fix: Add support for asymmetric error bars in deltakit.analysis

mhmohona:
## 🔗 Closed Issues

Link any issues closed or comments addressed by this PR.
"Closes #X" will close issue X automatically when the PR is merged.
Closes #134 

---

## 📝 Description
 
| asymmetric_lep_uncertainty_demo_deltakit_api | asymmetric_lep_uncertainty_demo |
|---|---|
 
Symmetric vs asymmetric: Gray = old ±3σ Gaussian; teal = binomial likelihood intervals (h=1000). At low rates, symmetric band is too narrow upward and can mislead near zero.

Deltakit plotting API: Data points use logical_error_probabilities_fit; orange line + shaded band = LEPPR fit (still symmetric on ε — known limitation).

Add any additional description required to understand the PR.
Cite related issues, PRs, or comments as needed.

---

## 🚦 Status

If anything still needs to be done before this PR can be merged, track it here.

---

## 🛠️ Future Work

Focused PRs are good!
Create and link issues for any follow-up work needed.

---

## ➕️ Additional Information

Anything that doesn't fit into a category above.

---

## 🧾 Release Note

If this PR will be **squash-merged**, draft the final one-line commit message here.
Otherwise, describe your rebase strategy or how you plan to maintain clean commits.

PR #275, opened 2026-06-06
feat(analysis): support asymmetric error bars

### LINKED ISSUE #134: request: support asymmetric error-bars in `deltakit.analysis`
### Motivation

For low logical error probabilities, symmetric error-bars may be imprecise as they artificially force a symmetric error even when the predicted value is close to 0.

It would be nice for the modules in `deltakit.analysis` to also support asymmetric error bars as that might improve the fits.

### Proposed solution

Start from [`sinter._probability_utils`](https://github.com/quantumlib/Stim/blob/main/glue/sample/src/sinter/_probability_util.py) code and [this explanation](https://quantumcomputing.stackexchange.com/a/37268/1386) of the problem for logical error probability.

See if we can re-use some of the already implemented methods in [`sinter._probability_utils`](https://github.com/quantumlib/Stim/blob/main/glue/sample/src/sinter/_probability_util.py), even if they are private. 

Ideally, we would like asymmetric error bars to be used and propagated everywhere. In practice, it might be hard to return correct asymmetric error bars when doing fitting, that needs more exploration.

### Potential alternatives

_No response_

### Additional context

> [!IMPORTANT]
> This issue is to be addressed in the broader context of #163 which aims at developing a `plotting` module to gather all plotting functionality under a single umbrella.

Ronit-Raj9:
## 🔗 Closed Issues

  Closes #134

  ---

  ## 📝 Description

  Symmetric error bars (`sqrt(p(1-p)/n)`) are a poor fit at low logical error
  rates: they can dip below zero and they understate the upper tail. This adds
  asymmetric error bars to `deltakit.analysis`, end to end:

  - **Data points:** binomial confidence intervals (`fit_binomial`,
    `calculate_lep_asymmetric`) that never go negative, ported from
    `sinter._probability_util`.
  - **Fits:** a binomial profile-likelihood fit of the per-round and SPAM errors
    (`compute_logical_error_per_round_asymmetric`) and of Λ
    (`calculate_lambda_asymmetric`), so the asymmetry comes from the fit itself,
    not just the bars. This is the chi-square = 1 / MINOS idea, using scipy only.
  - **Downstream:** the asymmetric bounds flow into the plotting bands, the LEPPR
    and Λ plot error bars, QuOps (`predict_quops_interval`), and the
    error-budget single-point `inverse_lambda_interval_at`.

  Existing functions are untouched — the asymmetric variants are opt-in siblings,
  so nothing downstream changes unless you ask for them. While wiring the plots up
  I also fixed a `TypeError` in `plot_logical_error_probability_per_round` and a Λ
  band that used `lambda0_std` for both Λ and Λ₀.

  I'm aware of #273, which covers the per-data-point binomial intervals; this PR
  additionally takes the asymmetry through the fits and the rest of the analysis.
  Happy to coordinate so we don't duplicate the sinter port.

  ---

  ## 📊 What it looks like

asym_demo
  
  *At low error rates the symmetric ±3σ bars cross below zero (left, unphysical); the 
  asymmetric binomial interval stays positive (right).*

interval_demo
 
  *Sweeping the number of observed failures: the symmetric Gaussian lower bound is negative
  for small counts, while the binomial interval stays in [0, 1]. The two converge once 

  ---

  ## 🚦 Status

  Ready for review. Locally green against the pinned tooling: `ruff`, `mypy`,
  `pydoclint`, and the full `analysis` + `plotting` test suites.

  ---

  ## 🛠️  Future Work

  - Propagate asymmetric Λ through the gradient-based `get_error_budget` (only the
    single-point `1/Λ` is covered here).
  - Share the binomial port with #273 once a home is agreed.

  ---

  ## ➕️ Additional Information

  `log_binomial` and `fit_binomial` are adapted from Stim's `sinter` (Apache-2.0),
  with attribution in the module. The per-round and SPAM errors are correlated, so
  on weakly constraining data the per-round interval can widen towards zero; pass
  `fixed_spam` if the SPAM error is known, and a warning is emitted otherwise.

  ---

  ## 🧾 Release Note

  feat(analysis): support asymmetric error bars in deltakit.analysis

### DIFF
--- a/deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
+++ b/deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 import dataclasses
 import math
 import warnings
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 
 import numpy as np
 import numpy.typing as npt
@@ -47,21 +47,8 @@ class ConfidenceInterval:
 
     def __post_init__(self) -> None:
         if not (self.low <= self.best <= self.high):
-            msg = (
-                f"Expected low <= best <= high, got "
-                f"{self.low}, {self.best}, {self.high}."
-            )
+            msg = f"Expected low <= best <= high, got {self.low}, {self.best}, {self.high}."
             raise ValueError(msg)
-
-    @property
-    def lower_margin(self) -> float:
-        """How far ``best`` sits above ``low``."""
-        return self.best - self.low
-
-    @property
-    def upper_margin(self) -> float:
-        """How far ``high`` sits above ``best``."""
-        return self.high - self.best
 
 
 def log_binomial(*, p: float, num_trials: int, num_successes: int) -> float:
@@ -98,9 +85,9 @@ def log_binomial(*, p: float, num_trials: int, num_successes: int) -> float:
 
 
 def fit_binomial(
-    *,
     num_shots: int,
     num_hits: int,
+    *,
     max_likelihood_factor: float = DEFAULT_MAX_LIKELIHOOD_FACTOR,
 ) -> ConfidenceInterval:
     """Estimate an error rate and its asymmetric confidence interval.
@@ -137,7 +124,11 @@ def fit_binomial(
     def gap(p: float) -> float:
         return log_binomial(p=p, num_trials=num_shots, num_successes=num_hits) - target
 
-    # The likelihood is unimodal, so each tail crosses the target exactly once.
+    # The binomial likelihood is unimodal in ``p`` with its peak at ``best``, so
+    # ``gap`` is positive at the peak and decreases monotonically towards each
+    # end. Each side therefore crosses zero exactly once, and a bracketed root
+    # finder (Brent's method) returns that single crossing robustly without
+    # needing derivatives.
     low = 0.0 if num_hits == 0 else float(optimize.brentq(gap, 1e-18, best))
     high = (
         1.0 if num_hits == num_shots else float(optimize.brentq(gap, best, 1 - 1e-18))
@@ -172,8 +163,8 @@ def fit_binomial_batch(
 
     fits = [
         fit_binomial(
-            num_shots=int(s),
-            num_hits=int(h),
+            int(s),
+            int(h),
             max_likelihood_factor=max_likelihood_factor,
         )
         for s, h in zip(shots.ravel(), hits.ravel(), strict=True)
@@ -193,11 +184,17 @@ def fit_binomial_batch(
 # to the raw counts by maximising the total binomial likelihood, then profile
 # each parameter to obtain an asymmetric interval. Profiling means fixing the
 # parameter of interest, re-optimising the other one, and walking outward until
-# the log-likelihood has dropped by ``num_sigmas ** 2 / 2`` (the chi-square = 1
-# rule for one sigma). This is the same idea as Minuit's MINOS errors and, unlike
-# a weighted least-squares fit with symmetric sigmas, it keeps the bounds inside
-# the physical range and reflects the genuine skew of the likelihood near zero.
+# the negative log-likelihood has risen by ``num_sigmas ** 2 / 2`` (so a 1-sigma
+# interval is reached at a rise of 0.5). This is the chi-square = 1 / MINOS rule
+# that follows from Wilks' theorem -- ``-2 ln(likelihood ratio)`` is
+# asymptotically chi-square distributed -- see
+# https://statproofbook.github.io/P/ci-wilks.html. Unlike a weighted
+# least-squares fit with symmetric sigmas, it keeps the bounds inside the
+# physical range and reflects the genuine skew of the likelihood near zero.
 
+# Physical bounds on a per-round/SPAM error probability: strictly above zero (so
+# the log-likelihood stays finite) and strictly below 0.5 (the maximally mixed
+# rate, where the model becomes degenerate).
 _RATE_FLOOR = 1e-12
 _RATE_CEIL = 0.5 - 1e-9
 
@@ -212,19 +209,22 @@ def _total_neg_log_likelihood(
     eps: float,
     spam: float,
     rounds: npt.NDArray[np.float64],
-    hits: npt.NDArray[np.int_],
+    num_fails: npt.NDArray[np.int_],
     shots: npt.NDArray[np.int_],
 ) -> float:
+    # The binomial "success" event here is a *logical failure*, so the per-point
+    # success count is the number of logical failures ``num_fails`` and ``p`` is
+    # the modelled logical error probability ``_model_lep``.
     model = _model_lep(rounds, eps, spam)
     return -sum(
         log_binomial(p=float(p), num_trials=int(n), num_successes=int(k))
-        for p, n, k in zip(model, shots, hits, strict=True)
+        for p, n, k in zip(model, shots, num_fails, strict=True)
     )
 
 
 def _best_fit(
     rounds: npt.NDArray[np.float64],
-    hits: npt.NDArray[np.int_],
+    num_fails: npt.NDArray[np.int_],
     shots: npt.NDArray[np.int_],
     eps_guess: float,
     spam_guess: float,
@@ -233,7 +233,7 @@ def _best_fit(
 
     Args:
         rounds: Number of QEC rounds for each point.
-        hits: Number of failures observed at each point.
+        num_fails: Number of logical failures observed at each point.
         shots: Number of shots at each point.
         eps_guess: Initial guess for the per-round error.
         spam_guess: Initial guess for the SPAM error.
@@ -245,8 +245,13 @@ def _best_fit(
     def cost(log_params: npt.NDArray[np.float64]) -> float:
         eps = float(np.clip(np.exp(log_params[0]), _RATE_FLOOR, _RATE_CEIL))
         spam = float(np.clip(np.exp(log_params[1]), _RATE_FLOOR, _RATE_CEIL))
-        return _total_neg_log_likelihood(eps, spam, rounds, hits, shots)
+        return _total_neg_log_likelihood(eps, spam, rounds, num_fails, shots)
 
+    # Nelder-Mead is derivative-free and robust for this smooth, low-dimensional
+    # cost; it is not guaranteed to find the global optimum, but the two rates
+    # are well scaled in log space and the subsequent profiling re-evaluates the
+    # likelihood around the returned point, so a small offset does not bias the
+    # reported interval.
     start = np.log([max(eps_guess, _RATE_FLOOR), max(spam_guess, _RATE_FLOOR)])
     result = optimize.minimize(cost, x0=start, method="Nelder-Mead")
     eps = float(np.clip(np.exp(result.x[0]), _RATE_FLOOR, _RATE_CEIL))
@@ -255,40 +260,29 @@ def _best_fit(
 
 
 def _profiled_neg_log_likelihood(
-    fixed_index: int,
+    fixed_eps: bool,
     fixed_value: float,
     rounds: npt.NDArray[np.float64],
-    hits: npt.NDArray[np.int_],
+    num_fails: npt.NDArray[np.int_],
     shots: npt.NDArray[np.int_],
-    fixed_other: float | None = None,
 ) -> float:
-    """Negative log-likelihood with one parameter fixed.
-
-    The other parameter is re-optimised unless ``fixed_other`` is given, in which
-    case it is held at that value (used when the SPAM error is known).
+    """Negative log-likelihood with one parameter fixed and the other refitted.
 
     Args:
-        fixed_index: 0 to fix the per-round error, 1 to fix the SPAM error.
+        fixed_eps: If True the per-round error is held at ``fixed_value`` and the
+            SPAM error is re-optimised; if False the roles are swapped.
         fixed_value: Value of the fixed parameter.
         rounds: Number of QEC rounds for each point.
-        hits: Number of failures observed at each point.
+        num_fails: Number of logical failures observed at each point.
         shots: Number of shots at each point.
-        fixed_other: Value to hold the other parameter at, or None to fit it.
 
     Returns:
-        The minimised negative log-likelihood.
+        The minimised negative log-likelihood over the free parameter.
     """
-    if fixed_other is not None:
-        eps, spam = (
-            (fixed_value, fixed_other)
-            if fixed_index == 0
-            else (fixed_other, fixed_value)
-        )
-        return _total_neg_log_likelihood(eps, spam, rounds, hits, shots)
 
     def cost(other: float) -> float:
-        eps, spam = (fixed_value, other) if fixed_index == 0 else (other, fixed_value)
-        return _total_neg_log_likelihood(eps, spam, rounds, hits, shots)
+        eps, spam = (fixed_value, other) if fixed_eps else (other, fixed_value)
+        return _total_neg_log_likelihood(eps, spam, rounds, num_fails, shots)
 
     result = optimize.minimize_scalar(
         cost, bounds=(_RATE_FLOOR, _RATE_CEIL), method="bounded"
@@ -296,29 +290,32 @@ def _profiled_neg_log_likelihood(
     return float(result.fun)
 
 
-def _profile_interval(
-    fixed_index: int,
+def _profile_from_cost(
+    cost: Callable[[float], float],
     best_value: float,
-    rounds: npt.NDArray[np.float64],
-    hits: npt.NDArray[np.int_],
-    shots: npt.NDArray[np.int_],
     num_sigmas: float,
-    fixed_other: float | None = None,
 ) -> ConfidenceInterval:
+    """Turn a profiled negative-log-likelihood into an asymmetric interval.
+
+    Walks outward from ``best_value`` until ``cost`` has risen by
+    ``num_sigmas ** 2 / 2`` above its value at the best fit (the chi-square = 1
+    rule), then locates that crossing on each side with a bracketed root find.
+
+    Args:
+        cost: The profiled negative log-likelihood as a function of the parameter
+            of interest, minimal at ``best_value``.
+        best_value: The maximum-likelihood value of the parameter.
+        num_sigmas: Width of the interval, in sigmas.
+
+    Returns:
+        The ``(low, best, high)`` confidence interval.
+    """
     # Anchor the target to the profiled likelihood at the best value (the global
     # minimum), so the root is always bracketed regardless of optimiser noise.
-    nll_at_best = _profiled_neg_log_likelihood(
-        fixed_index, best_value, rounds, hits, shots, fixed_other
-    )
-    target = nll_at_best + 0.5 * num_sigmas**2
+    target = cost(best_value) + 0.5 * num_sigmas**2
 
     def excess(value: float) -> float:
-        return (
-            _profiled_neg_log_likelihood(
-                fixed_index, value, rounds, hits, shots, fixed_other
-            )
-            - target
-        )
+        return cost(value) - target
 
     def cross(bound: float) -> float:
         # If the likelihood never rises enough before the bound, the parameter is
@@ -335,6 +332,35 @@ def _profile_interval(
     )
 
 
+def _profile_interval(
+    fixed_eps: bool,
+    best_value: float,
+    rounds: npt.NDArray[np.float64],
+    num_fails: npt.NDArray[np.int_],
+    shots: npt.NDArray[np.int_],
+    num_sigmas: float,
+) -> ConfidenceInterval:
+    """Profile one parameter (refitting the other) for an asymmetric interval.
+
+    Args:
+        fixed_eps: If True profile the per-round error (refitting SPAM); if False
+            profile the SPAM error (refitting the per-round error).
+        best_value: The maximum-likelihood value of the profiled parameter.
+        rounds: Number of QEC rounds for each point.
+        num_fails: Number of logical failures observed at each point.
+        shots: Number of shots at each point.
+        num_sigmas: Width of the interval, in sigmas.
+
+    Returns:
+        The ``(low, best, high)`` confidence interval for the profiled parameter.
+    """
+
+    def cost(value: float) -> float:
+        return _profiled_neg_log_likelihood(fixed_eps, value, rounds, num_fails, shots)
+
+    return _profile_from_cost(cost, best_value, num_sigmas)
+
+
 def fit_leppr_and_spam(
     num_rounds: npt.NDArray[np.int_] | Sequence[int],
     num_fails: npt.NDArray[np.int_] | Sequence[int],
@@ -345,10 +371,10 @@ def fit_leppr_and_spam(
 ) -> tuple[ConfidenceInterval, ConfidenceInterval]:
     """Fit the per-round error and SPAM error with asymmetric intervals.
 
-    The model ``LEP(r) = (1 - (1 - 2 spam)(1 - 2 eps) ** r) / 2`` is fitted to the
-    raw counts by maximum binomial likelihood, and each parameter is profiled to
-    get an asymmetric confidence interval. ``num_sigmas = 1`` corresponds to a
-    chi-square = 1 interval.
+    The model ``LEP(r) = (1 - (1 - 2 * spam) * (1 - 2 * eps) ** r) / 2`` is fitted
+    to the raw counts by maximum binomial likelihood, and each parameter is
+    profiled to get an asymmetric confidence interval. ``num_sigmas = 1``
+    corresponds to a chi-square = 1 interval.
 
     The per-round error and the SPAM error are correlated, so when the data only
     weakly constrains the per-round error its interval can extend down to zero. If
@@ -397,16 +423,20 @@ def fit_leppr_and_spam(
             method="bounded",
         )
         eps_best = float(np.clip(np.exp(eps_only.x), _RATE_FLOOR, _RATE_CEIL))
-        leppr_fit = _profile_interval(
-            0, eps_best, rounds, fails, shots, num_sigmas, fixed_other=spam_best
+        # SPAM is held fixed, so the per-round profile is just the total
+        # likelihood as a function of eps alone.
+        leppr_fit = _profile_from_cost(
+            lambda eps: _total_neg_log_likelihood(eps, spam_best, rounds, fails, shots),
+            eps_best,
+            num_sigmas,
         )
         spam_fit = ConfidenceInterval(low=spam_best, best=spam_best, high=spam_best)
         return leppr_fit, spam_fit
 
     spam_guess = float(np.clip(lep[0] if lep.size else 0.0, _RATE_FLOOR, _RATE_CEIL))
     eps_best, spam_best = _best_fit(rounds, fails, shots, eps_guess, spam_guess)
-    leppr_fit = _profile_interval(0, eps_best, rounds, fails, shots, num_sigmas)
-    spam_fit = _profile_interval(1, spam_best, rounds, fails, shots, num_sigmas)
+    leppr_fit = _profile_interval(True, eps_best, rounds, fails, shots, num_sigmas)
+    spam_fit = _profile_interval(False, spam_best, rounds, fails, shots, num_sigmas)
 
     if leppr_fit.low <= _RATE_FLOOR * 10 < leppr_fit.best:
         warnings.warn(
--- a/deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
+++ b/deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
@@ -18,6 +18,7 @@ from __future__ import annotations
 
 import dataclasses
 import math
+import warnings
 from collections.abc import Sequence
 
 import numpy as np
@@ -236,8 +237,20 @@ def _profiled_neg_log_likelihood(
     rounds: npt.NDArray[np.float64],
     hits: npt.NDArray[np.int_],
     shots: npt.NDArray[np.int_],
+    fixed_other: float | None = None,
 ) -> float:
-    """Negative log-likelihood with one parameter fixed and the other free."""
+    """Negative log-likelihood with one parameter fixed.
+
+    The other parameter is re-optimised unless ``fixed_other`` is given, in which
+    case it is held at that value (used when the SPAM error is known).
+    """
+    if fixed_other is not None:
+        eps, spam = (
+            (fixed_value, fixed_other)
+            if fixed_index == 0
+            else (fixed_other, fixed_value)
+        )
+        return _total_neg_log_likelihood(eps, spam, rounds, hits, shots)
 
     def cost(other: float) -> float:
         eps, spam = (fixed_value, other) if fixed_index == 0 else (other, fixed_value)
@@ -256,17 +269,20 @@ def _profile_interval(
     hits: npt.NDArray[np.int_],
     shots: npt.NDArray[np.int_],
     num_sigmas: float,
+    fixed_other: float | None = None,
 ) -> Fit:
     # Anchor the target to the profiled likelihood at the best value (the global
     # minimum), so the root is always bracketed regardless of optimiser noise.
     nll_at_best = _profiled_neg_log_likelihood(
-        fixed_index, best_value, rounds, hits, shots
+        fixed_index, best_value, rounds, hits, shots, fixed_other
     )
     target = nll_at_best + 0.5 * num_sigmas**2
 
     def excess(value: float) -> float:
         return (
-            _profiled_neg_log_likelihood(fixed_index, value, rounds, hits, shots)
+            _profiled_neg_log_likelihood(
+                fixed_index, value, rounds, hits, shots, fixed_other
+            )
             - target
         )
 
@@ -289,6 +305,7 @@ def fit_leppr_and_spam(
     num_shots: npt.NDArray[np.int_] | Sequence[int],
     *,
     num_sigmas: float = 1.0,
+    fixed_spam: float | None = None,
 ) -> tuple[Fit, Fit]:
     """Fit the per-round error and SPAM error with asymmetric intervals.
 
@@ -297,11 +314,19 @@ def fit_leppr_and_spam(
     get an asymmetric confidence interval. ``num_sigmas = 1`` corresponds to a
     chi-square = 1 interval.
 
+    The per-round error and the SPAM error are correlated, so when the data only
+    weakly constrains the per-round error its interval can extend down to zero. If
+    the SPAM error is known independently, pass it as ``fixed_spam`` to remove the
+    correlation and obtain a tighter per-round interval. A warning is emitted when
+    the per-round interval is left open towards zero.
+
     Args:
         num_rounds: Number of QEC rounds for each measured point.
         num_fails: Number of logical failures observed at each round count.
         num_shots: Number of shots at each round count.
         num_sigmas: Width of the interval, in sigmas.
+        fixed_spam: A known SPAM error to hold fixed. When given, only the
+            per-round error is fitted.
 
     Returns:
         ``(leppr_fit, spam_fit)`` as :class:`Fit` instances.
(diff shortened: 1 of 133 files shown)


deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,404 @@
...
+from collections.abc import Sequence
+
+import numpy as np
+import numpy.typing as npt
+from scipy import optimize
+
+#: Default Bayes factor used to define the confidence interval. Hypotheses whose
+#: likelihood is more than this many times less likely than the best fit are
+#: excluded. A factor of 1000 corresponds to roughly a 99% interval.
+DEFAULT_MAX_LIKELIHOOD_FACTOR = 1000.0
```
rolandriver:
```suggestion
DEFAULT_MAX_LIKELIHOOD_FACTOR: float = 1000.0
```

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+@dataclasses.dataclass(frozen=True)
+class Fit:
```
rolandriver: Fit is too vague of a class name. Any attempt for a more meaningful one?

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,404 @@
...
+    """A point estimate together with its lower and upper bounds.
+
+    Attributes:
+        low: Smallest rate compatible with the data at the chosen confidence.
+        best: Maximum-likelihood estimate (``num_hits / num_shots``).
+        high: Largest rate compatible with the data at the chosen confidence.
+    """
+
+    low: float
```
rolandriver, line 44: I would check for values in the [0,1] range.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,404 @@
...
+    best: float
+    high: float
+
+    @property
+    def lower_margin(self) -> float:
+        """How far ``best`` sits above ``low``."""
+        return self.best - self.low
```
rolandriver: Is it always a positive number?

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,404 @@
...
+
+    @property
+    def upper_margin(self) -> float:
+        """How far ``high`` sits above ``best``."""
+        return self.high - self.best
```
rolandriver: Same.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,404 @@
...
+
+
+def log_binomial(*, p: float, n: int, hits: int) -> float:
```
rolandriver:
```suggestion
def log_binomial(*, p: float, num_trials: int, num_successes: int) -> float:
```

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,404 @@
...
+        n: Number of shots.
+        hits: Number of failures observed.
+
+    Returns:
+        The natural log of the binomial likelihood.
+    """
+    p = min(max(p, 0.0), 1.0)
+    misses = n - hits
+    result = 0.0
+    if hits:
```
rolandriver:
```suggestion
    if hits != 0:
```

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,404 @@
...
+        if p == 0:
+            return -math.inf
+        result += math.log(p) * hits
+    if misses:
```
rolandriver:
```suggestion
    if misses != 0:
```

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
```diff
@@ -379,3 +394,167 @@ def calculate_lambda_and_lambda_stddev(
...
+    )
+
+    # Λ = exp(-slope) is decreasing in the slope, so the bounds swap over.
+    lambda_low = float(np.exp(-slope_fit.high))
+    lambda_high = float(np.exp(-slope_fit.low))
+    lambda0_low = float(np.exp(-offset_fit.high))
+    lambda0_high = float(np.exp(-offset_fit.low))
+
+    return LambdaData(
+        lambda_=central.lambda_,
```
rolandriver, line 651: This becomes a bit cluttered. I would think about custom data types to group and hold related data.

rolandriver (commented):
Thanks again @Ronit-Raj9 for contributing to Deltakit. This looks quite nice to me. As you are aware, there are other contributions to this issue. Please bear with us while we discuss with other maintainers. Thank you.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Sure, i will put some better meaningful name

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: thanks for this.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Renamed to ConfidenceInterval - matches the *_interval helpers that return it.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: `Fit` isn't rate-only here - `_profile_scalar` returns it for the slope/offset (can be negative), predict_quops_interval` for a QuOps count (way above 1),  and  `inverse_lambda_interval_at` for 1/Λ. So a [0,1] check would reject valid fits. I added a `low <= best <= high` invariant in `__post_init__` instead, which also makes the two margins non-negative by construction (your next two comments).

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Yes - low <= best <= high holds at every construction site, and __post_init__ now enforces it, so best - low >= 0 always.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Same - guaranteed by the invariant, so high - best >= 0.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: cool

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
Ronit-Raj9: Good point. I've introduced a `ConfidenceInterval` type and used it for the fit returns. For `LambdaData` itself, `lambda_low/high`, `lambda0_low/high` and `leppr_low/high` are public fields the plotting layer reads directly, so folding them into a grouped type reshapes a public dataclass and its readers. Happy to do that - just wanted to check you'd prefer it over the flat fields before changing the public type. Let me know which way you'd like.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Done

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: did it

rolandriver:
@Ronit-Raj9 Could you please post examples of plotting before and after that you can generate with this? It'll help making a decision. Thank you.

Ronit-Raj9:
> @Ronit-Raj9 Could you please post examples of plotting before and after that you can generate with this? It'll help making a decision. Thank you.

yupp... i will post in a while and also fix these build errors too...

Ronit-Raj9:
Hi @rolandriver , I have updated PR body with image... You can check it.... Going to fix CI issue... Lets me know if anything else is needed

Ronit-Raj9:
Done with CI fixes.. You can review it finally...  Images have been added too in the PR body

deltakit-explorer/src/deltakit_explorer/analysis/error_budget/_lambda.py
```diff
@@ -70,40 +127,67 @@ def inverse_lambda_at(
...
+        memory_generator,
     )
-    report = engine.run()
     lambdas, lambda_stddevs = compute_lambda_and_stddev_from_results(
         point, noise_parameter_names, num_rounds_by_distances, report
     )
     lambda_reciprocals = 1 / lambdas
     lambda_reciprocal_stddevs = np.vectorize(reciprocal_stddev)(lambdas, lambda_stddevs)
 
     return float(lambda_reciprocals[0, 0]), float(lambda_reciprocal_stddevs[0, 0])
```
rolandriver: Why don't you return a `ConfidenceInterval` here?

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+    best = num_hits / num_shots
+    target = log_binomial(
+        p=best, num_trials=num_shots, num_successes=num_hits
+    ) - math.log(max_likelihood_factor)
+
+    def gap(p: float) -> float:
+        return log_binomial(p=p, num_trials=num_shots, num_successes=num_hits) - target
+
+    # The likelihood is unimodal, so each tail crosses the target exactly once.
+    low = 0.0 if num_hits == 0 else float(optimize.brentq(gap, 1e-18, best))
```
rolandriver, line 132: Could you please explain why choosing this method?

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+        The maximum-likelihood ``(eps, spam)``.
+    """
+
+    def cost(log_params: npt.NDArray[np.float64]) -> float:
+        eps = float(np.clip(np.exp(log_params[0]), _RATE_FLOOR, _RATE_CEIL))
+        spam = float(np.clip(np.exp(log_params[1]), _RATE_FLOOR, _RATE_CEIL))
+        return _total_neg_log_likelihood(eps, spam, rounds, hits, shots)
+
+    start = np.log([max(eps_guess, _RATE_FLOOR), max(spam_guess, _RATE_FLOOR)])
+    result = optimize.minimize(cost, x0=start, method="Nelder-Mead")
```
rolandriver, line 256: Is this guaranteed to converge?

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+    eps = float(np.clip(np.exp(result.x[0]), _RATE_FLOOR, _RATE_CEIL))
+    spam = float(np.clip(np.exp(result.x[1]), _RATE_FLOOR, _RATE_CEIL))
+    return eps, spam
+
+
+def _profiled_neg_log_likelihood(
+    fixed_index: int,
```
rolandriver: Shouldn't that be a `bool` ? In any case, it feels quite fragile as there is no check for this parameter to be of the right values. I would try a more solid approach based on enums.

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
```diff
@@ -120,6 +122,13 @@ class LambdaData:
         distances: An array of code distances.
         leppr: An array for leppr computed for all code distances.
         leppr_std: An array for leppr standard deviation computed for all code distances.
+        lambda_low: Lower bound of Λ. Only set by
```
rolandriver: Again, I would reduce this or think about a way to pass these confidence intervals as it reduces the usability of `LambdaData` IMHO.

deltakit-explorer/src/deltakit_explorer/analysis/_leppr.py
```diff
@@ -447,6 +462,60 @@ def perfect_simulator(num_rounds: int) -> tuple[int, int]:
     return np.asarray(nrounds), np.asarray(nfails), np.asarray(nshots)
 
 
+def compute_logical_error_per_round_asymmetric(
```
rolandriver: I would rename it as I get confused with the `calculate_lep_asymmetric` below.

deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
```diff
@@ -83,6 +86,48 @@ def predict_quops_at_distance(lambda0: float, lambda_: float, distance: int) ->
     return 1.0 / _calculate_lep(lambda0, lambda_, distance, distance)
 
 
+def predict_quops_interval(
+    lambda_data: LambdaData, distance: int
+) -> ConfidenceInterval:
+    """Predict the number of QuOps with a confidence interval.
```
rolandriver, line 92: What do you mean by `QuOps` exactly?

rolandriver (commented):
Thanks @Ronit-Raj9 some more comments from me.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+    low: float
+    best: float
+    high: float
+
+    def __post_init__(self) -> None:
+        if not (self.low <= self.best <= self.high):
+            msg = (
+                f"Expected low <= best <= high, got "
+                f"{self.low}, {self.best}, {self.high}."
+            )
```
nelimee:
```suggestion
            msg = f"Expected low <= best <= high, got {self.low}, {self.best}, {self.high}."

```

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+    @property
+    def lower_margin(self) -> float:
+        """How far ``best`` sits above ``low``."""
+        return self.best - self.low
+
+    @property
+    def upper_margin(self) -> float:
+        """How far ``high`` sits above ``best``."""
+        return self.high - self.best
+
```
nelimee: These two are only used in tests. Seems like unused properties to me, maybe remove them?

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+        1.0 if num_hits == num_shots else float(optimize.brentq(gap, best, 1 - 1e-18))
+    )
+    return ConfidenceInterval(low=low, best=best, high=high)
+
+
+def fit_binomial_batch(
+    num_shots: npt.NDArray[np.int_] | Sequence[int],
+    num_hits: npt.NDArray[np.int_] | Sequence[int],
+    *,
+    max_likelihood_factor: float = DEFAULT_MAX_LIKELIHOOD_FACTOR,
```
nelimee, line 143: Not the same signature as `fit_binomial` (`*` is at another place), would probably benefit a uniformisation.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+# to the raw counts by maximising the total binomial likelihood, then profile
+# each parameter to obtain an asymmetric interval. Profiling means fixing the
+# parameter of interest, re-optimising the other one, and walking outward until
+# the log-likelihood has dropped by ``num_sigmas ** 2 / 2`` (the chi-square = 1
+# rule for one sigma). This is the same idea as Minuit's MINOS errors and, unlike
+# a weighted least-squares fit with symmetric sigmas, it keeps the bounds inside
+# the physical range and reflects the genuine skew of the likelihood near zero.
+
+_RATE_FLOOR = 1e-12
+_RATE_CEIL = 0.5 - 1e-9
```
nelimee, line 199: Could use a small comment to explain what these variables are used for.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
nelimee: I think you are right, this parameter should likely be something like `fixed_eps: bool`.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+    Returns:
+        The minimised negative log-likelihood.
+    """
+    if fixed_other is not None:
+        eps, spam = (
+            (fixed_value, fixed_other)
+            if fixed_index == 0
+            else (fixed_other, fixed_value)
+        )
+        return _total_neg_log_likelihood(eps, spam, rounds, hits, shots)
```
nelimee: I feel like `fixed_other` is complexifying that function for very little gains. Why not calling `_total_neg_log_likelihood` directly at the only location where `fixed_other` is set?

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+
+def _profile_interval(
+    fixed_index: int,
+    best_value: float,
+    rounds: npt.NDArray[np.float64],
+    hits: npt.NDArray[np.int_],
+    shots: npt.NDArray[np.int_],
+    num_sigmas: float,
+    fixed_other: float | None = None,
+) -> ConfidenceInterval:
```
nelimee: Would probably benefit from having a docstring

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+    num_rounds: npt.NDArray[np.int_] | Sequence[int],
+    num_fails: npt.NDArray[np.int_] | Sequence[int],
+    num_shots: npt.NDArray[np.int_] | Sequence[int],
+    *,
+    num_sigmas: float = 1.0,
+    fixed_spam: float | None = None,
+) -> tuple[ConfidenceInterval, ConfidenceInterval]:
+    """Fit the per-round error and SPAM error with asymmetric intervals.
+
+    The model ``LEP(r) = (1 - (1 - 2 spam)(1 - 2 eps) ** r) / 2`` is fitted to the
```
nelimee:
```suggestion
    The model ``LEP(r) = (1 - (1 - 2 spam) * (1 - 2 eps) ** r) / 2`` is fitted to the
```

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
```diff
@@ -0,0 +1,418 @@
...
+
+    if fixed_spam is not None:
+        spam_best = float(np.clip(fixed_spam, _RATE_FLOOR, _RATE_CEIL))
+        # Fit in log space so the small rates are well scaled.
+        eps_only = optimize.minimize_scalar(
+            lambda log_eps: _total_neg_log_likelihood(
+                float(np.clip(np.exp(log_eps), _RATE_FLOOR, _RATE_CEIL)),
+                spam_best,
+                rounds,
+                fails,
```
nelimee, line 419: Why is that `fails` and not `shots - fails`? This parameter is supposed to be the `hits` no?

deltakit-explorer/src/deltakit_explorer/analysis/_leppr.py
```diff
@@ -78,6 +78,12 @@ def epsilon_and_spam_from_log_fit(
         Estimate.from_ufloat(uncertain_spam),
     )
 
+from deltakit_explorer.analysis._binomial_fit import (
+    DEFAULT_MAX_LIKELIHOOD_FACTOR,
+    fit_binomial_batch,
+    fit_leppr_and_spam,
+)
```
nelimee: Ruff emits `E402` here.

deltakit-explorer/src/deltakit_explorer/analysis/_leppr.py
```diff
@@ -447,6 +462,60 @@ def perfect_simulator(num_rounds: int) -> tuple[int, int]:
...
+    shots = np.asarray(num_shots)
+    order = np.argsort(rounds)
+    rounds, fails, shots = rounds[order], fails[order], shots[order]
+
+    leppr_fit, spam_fit = fit_leppr_and_spam(
+        rounds, fails, shots, num_sigmas=num_sigmas, fixed_spam=fixed_spam
+    )
+    return LogicalErrorProbabilityPerRoundData(
+        leppr=leppr_fit.best,
+        leppr_stddev=(leppr_fit.high - leppr_fit.low) / 2,
```
nelimee, line 513: Could you explain where that formula comes from?

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
```diff
@@ -93,6 +93,8 @@ def lambda_from_curve_fit(
         Estimate.from_ufloat(uncertain_lamb0),
     )
 
+from deltakit_explorer.analysis._binomial_fit import ConfidenceInterval
+
```
nelimee: Ruff `E402`

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
```diff
@@ -457,3 +472,167 @@ def calculate_lambda_and_lambda_stddev(
...
+    Args:
+        cost: A function of one parameter, minimal at ``best``.
+        best: The best-fit value of the parameter.
+        num_sigmas: Width of the interval, in sigmas.
+
+    Returns:
+        The best value and its lower and upper bounds.
+    """
+    target = cost(best) + 0.5 * num_sigmas**2
```
nelimee, line 496: Could you explain that formula or link me to a place where it is explained?

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
```diff
@@ -457,3 +472,167 @@ def calculate_lambda_and_lambda_stddev(
...
+
+    def excess(value: float) -> float:
+        return cost(value) - target
+
+    def walk(direction: int) -> float:
```
nelimee: Really nitpick for this one.
```suggestion
    def walk(direction: Literal[-1, 1]) -> float:
```

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
```diff
@@ -457,3 +472,167 @@ def calculate_lambda_and_lambda_stddev(
...
+        num_sigmas: Width of the interval, in sigmas.
+
+    Returns:
+        The ``(slope, offset)`` fits, each with its lower and upper bounds.
+    """
+
+    def cost(slope: float, offset: float) -> float:
+        residual = offset + slope * x - y
+        sigma = np.where(residual >= 0, sigma_high, sigma_low)
+        return 0.5 * float(np.sum((residual / sigma) ** 2))
```
nelimee, line 546: Could you explain that formula (or point me to somewhere that explains it)?

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
```diff
@@ -457,3 +472,167 @@ def calculate_lambda_and_lambda_stddev(
...
+
+    def objective(params: npt.NDArray[np.float64]) -> float:
+        return cost(float(params[0]), float(params[1]))
+
+    slope0, offset0 = np.polyfit(x, y, 1)
```
nelimee: Use the newer https://numpy.org/doc/stable/reference/generated/numpy.polynomial.polynomial.Polynomial.fit.html#numpy.polynomial.polynomial.Polynomial.fit instead.

deltakit-explorer/src/deltakit_explorer/analysis/error_budget/_lambda.py
```diff
@@ -70,40 +127,67 @@ def inverse_lambda_at(
...
+    point, noise_parameter_names, report = _run_lambda_engine(
+        noise_model,
+        noise_parameters,
+        num_rounds_by_distances,
+        sampling_parameters,
+        memory_generator,
+    )
+    filtered = _filter_non_close_noise_parameters(
+        report, point[:, 0], noise_parameter_names
+    )
```
nelimee, line 192: I am surprised to see that line here and not in the symmetric version. Why did you need it?

deltakit-explorer/src/deltakit_explorer/plotting/results.py
```diff
@@ -188,20 +241,43 @@ def interpolate_lambda(
...
+    # give the lower boundary of the band and the smallest give the upper one.
+    interpolated = _lambda_interpolated(
+        lambda_data.lambda0, lambda_data.lambda_, distances_interpolated
+    ).ravel()
+    lower_boundary = _lambda_interpolated(
+        lambda0_high, lambda_high, distances_interpolated
+    ).ravel()
+    upper_boundary = _lambda_interpolated(
+        lambda0_low, lambda_low, distances_interpolated
+    ).ravel()
```
nelimee, line 276: Numpy type errors here.

deltakit-explorer/src/deltakit_explorer/plotting/results.py
```diff
@@ -188,20 +241,43 @@ def interpolate_lambda(
...
-    # Offsets
-    offsets = np.array([-num_sigmas, 0.0, num_sigmas]) * lambda_data.lambda0_std
+    # Use the bounds from an asymmetric fit if they are available, otherwise
+    # build a multiplicative band from the symmetric standard deviations (Λ and Λ₀
+    # are exponentials of the fit parameters, so the band is not symmetric).
+    if (
+        lambda_data.lambda_low is not None
+        and lambda_data.lambda_high is not None
+        and lambda_data.lambda0_low is not None
+        and lambda_data.lambda0_high is not None
```
nelimee: I think you used that more than once. May be worth defining a `@property` on the data structure to avoid repeated this long condition.

nelimee (commented):
At first glance, this PR seems very good! I would love to have links / more explanations on some of the maths, but overall I think this is solid.

I'll re-do a review on the maths tomorrow, hopefully with a few more comments explaining where the formulas come from.

rolandriver:
Hi @Ronit-Raj9 UnitaryHack is due to end tomorrow. Will you have time to implement review comments and suggestions?

Ronit-Raj9:
Hi @rolandriver, sorry I was traveling a bit... Yess I am going to implement these changes...

deltakit-explorer/src/deltakit_explorer/analysis/error_budget/_lambda.py
Ronit-Raj9: Done. It returns a ConfidenceInterval now (symmetric value +/- stddev). Keeps it in line with inverse_lambda_interval_at, which gives the asymmetric one.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: The binomial likelihood is unimodal, so each tail crosses the threshold exactly once. That's a single bracketed root, which is what Brent's method handles well, no derivatives needed. Added a comment.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Not globally guaranteed. But it's a smooth 2-D cost, well scaled in log space, and the profiling step re-evaluates the likelihood around the result, so a small offset doesn't bias the interval. Added a note.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Agreed, it's fixed_eps: bool now.

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
Ronit-Raj9: Done. Folded the four Lambda/Lambda0 bound floats into lambda_interval / lambda0_interval ConfidenceInterval), and ropped the per-distance leppr bounds since nothing read them. Added a 
has_asymmetric_bounds property too.

deltakit-explorer/src/deltakit_explorer/analysis/_leppr.py
Ronit-Raj9: enamed to fit_logical_error_per_round_asymmetric. The fit_ vs calculate_ split makes the difference clearer.

deltakit-explorer/src/deltakit_explorer/analysis/_quops.py
Ronit-Raj9: Expanded the docstring. QuOps is the number of logical operations you can run before a logical error, taken as 1/pL for a distance-D, D-round block.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: done

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Removed. The tests just use high - best / best - low directly now.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Done. Both are (num_shots, num_hits, *, max_likelihood_factor) now.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Added. They're the physical bounds on a rate, just above 0 and just below 0.5.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Yupp fixed this. Dropped fixed_other. The fixed-spam branch now calls _total_neg_log_likelihood directly through a small shared _profile_from_cost helper.

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: Added

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: done

deltakit-explorer/src/deltakit_explorer/analysis/_binomial_fit.py
Ronit-Raj9: The binomial "success" event here is a logical failure (p = the logical error rate), so num_successes is fails. Renamed the internal param hits to num_fails and added a comment so it's not confusing.

deltakit-explorer/src/deltakit_explorer/analysis/_leppr.py
Ronit-Raj9: Fixed, moved to the top.

deltakit-explorer/src/deltakit_explorer/analysis/_leppr.py
Ronit-Raj9: It's the symmetric-equivalent sigma, half the asymmetric interval width, so anything expecting a single stddev still works. Added a comment.

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
Ronit-Raj9: fixed this one

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
Ronit-Raj9: It's the profile-likelihood / MINOS rule (Wilks' theorem): a 1-sigma interval is where the negative log-likelihood rises by 0.5. Added a comment and link: https://statproofbook.github.io/P/ci-wilks.html

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
Ronit-Raj9: done

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
Ronit-Raj9: Asymmetric chi-square: each residual is scaled by the upper error above the point and the lower error below. Added a comment and ref - Barlow, "Asymmetric Statistical Errors" (arXiv:physics/0406120).

deltakit-explorer/src/deltakit_explorer/analysis/_lambda.py
Ronit-Raj9: Switched here for the initial guess. Left np.polyfit in the LIN/SHIFTED fits since they need cov='unscaled', which Polynomial.fit doesn't expose. There's a comment there.

deltakit-explorer/src/deltakit_explorer/analysis/error_budget/_lambda.py
Ronit-Raj9: The symmetric compute_lambda_and_stddev_from_results filters inside its per-point loop. This one is single-point, so the caller selects the point first. Added a comment.

deltakit-explorer/src/deltakit_explorer/plotting/results.py
Ronit-Raj9: Fixed. Widened _lambda_interpolated to float | NDArray, since it's called with scalar Lambda/Lambda0 against an array of distances.

deltakit-explorer/src/deltakit_explorer/plotting/results.py
Ronit-Raj9: Done, that's the has_asymmetric_bounds property, used in both spots.

Ronit-Raj9:
CI is green now. The highest collection failures were pytest 9.1 turning a new deprecation (passing a bare itertools iterator to @parametrize) into an error under the repo's filterwarnings=['error']. It's repo-wide (deltakit-circuit tests too), surfaced by the pytest bump in #293. I added a scoped ignore for just that warning in the root pytest config, next to the "temporary fix" ignore already there. Happy to wrap the iterators in list(...) across those tests instead if you'd rather go that way.

nelimee (approved):
To me, this is a very nice addition that solves the issue. I am sorry for being a little bit slow reviewing, I was at a conference for work. 

There are some maths I would like to explore more in this PR but:

- I will have the time soon,
- I understand most of it, and the parts I am less familiar with are well commented and include links to explain the approach, which I skimmed through and make me confident this makes sense,
- Unitary Hack deadline was yesterday,

so I am OK to merge. 

Well done!

rolandriver (approved):
Looking good to me. Well done @Ronit-Raj9.

Ronit-Raj9:
Thanks @rolandriver and @nelimee.... Do let me know if any changes or updation is needed i will do that

rolandriver:
@nelimee Could you please close any pending comments so that I can merge? Thank you 🙏

nelimee:
> @nelimee Could you please close any pending comments so that I can merge? Thank you 🙏

Done, I ended up closing some of yours too. All comments are now closed.

Issue #248, opened 2026-04-20
request: Remove stim dependency in favour of deltakit-stim


rolandriver:
### Contributor License Agreement

- [x] A Contributor License Agreement signature is requested for any first PR contribution. Further information can be found in our [contributing guidelines](CONTRIBUTING.md).

### Motivation

Remove `stim` dependency in favour of `deltakit-stim`.

This encompasses other issues:
 - #247 
 - #152 

### Proposed solution

Replace `stim` imports and references with `deltakit-stim`.

rolandriver:
Linked to #247

PR #296, opened 2026-06-16
fix(dependencies): deltakit-stim>=0.2

nelimee:
## 🔗 Closed Issues

Require the new version of `deltakit-stim`, solving any interaction issue with `stim`.

---

## 📝 Description

This PR bumps `deltakit-stim` to `>=0.2` which is the first version with correct `stim` support https://github.com/Deltakit/deltakit-stim/releases/tag/v0.2.0.

---

## 🚦 Status

Ready to merge

---

## 🛠️ Future Work

None

---

## ➕️ Additional Information

None

---

## 🧾 Release Note

fix(dependencies): deltakit-stim>=0.2

PR #277, opened 2026-06-10
feat: Replace stim imports

rolandriver:
## 🔗 Closed Issues

Closes #248, #278 and partially #247.

---

## 📝 Description

This PR addresses the replacement of all `stim` imports by `deltakit-stim`.

---

## 🚦 Status

In progress.

---

## 🛠️ Future Work

NA.

---

## ➕️ Additional Information

NA.

---

## 🧾 Release Note

NA.

PR #298, opened 2026-06-16
fix(circuit): Circuit propagate tags as stim circuit 262

### LINKED ISSUE #262: bug: Method `as_stim_circuit` in `deltakit_circuit.Circuit` does not propagate tags.
### Contributor License Agreement

- [x] A Contributor Licence Agreement signature is required for any first PR contribution. Further information can be found in our [contributing guidelines](CONTRIBUTING.md).

### Bug Description

While running test `test_tag_parsing.py` in `deltakit_circuit`, the construction of a `deltakit_circuit.Circuit` instance gets the tag field correctly but it is omitted at export to a `stim` circuit:

```sh
(Pdb) circuit
Circuit([
    GateLayer([
        SQRT_Y_DAG[sjkdhf](Qubit(0))
    ])
], iterations=1)
(Pdb) stim_circuit
stim.Circuit('''
    SQRT_Y_DAG[sjkdhf] 0
''')
(Pdb) circuit.as_stim_circuit()
stim.Circuit('''
    SQRT_Y_DAG 0
''')
```

### Minimal Reproducible Example

Run `test_tag_parsing.py` in `deltakit_circuit`. Example in the previous section.

### Actual behaviour

The tag is not propagated at export time to compare stim circuits faithfully.

### Expected behaviour

The tag must be propagated at export time to compare stim circuits faithfully.

### Additional context

_No response_

lukedev45:
## Closed Issues

Closes #262

---

## Description

`Circuit.as_stim_circuit()` silently dropped gate **tags**. A circuit built from
`SQRT_Y_DAG[sjkdhf] 0` exported back as `SQRT_Y_DAG 0`.

Root cause is in `_stim_version_compatibility.py`: `is_stim_tag_feature_available()`
returns `False` whenever `importlib.metadata.version("stim")` raises
`PackageNotFoundError`. That happens when the importable `stim` module is provided
by a distribution **not** named `stim` (notably `deltakit-stim`), or when
distribution metadata is otherwise unreadable (editable / vendored installs).
`_gate_layer.py` only forwards the `tag=` keyword to stim when the feature is
"available", so in those environments the tag is dropped on export.

This PR makes `_get_stim_version()` robust:
- it also looks up the `deltakit-stim` distribution, and
- falls back to the imported `stim` module's `__version__` when distribution
  metadata can't be resolved.

The tag feature is now enabled whenever a tag-capable stim (≥ 1.15) is actually
importable, regardless of how it was packaged.

Issue repro, before vs after:

```python
>>> Circuit.from_stim_circuit(stim.Circuit("SQRT_Y_DAG[sjkdhf] 0")).as_stim_circuit()
# before:
stim.Circuit('SQRT_Y_DAG 0')          # tag dropped
# after:
stim.Circuit('SQRT_Y_DAG[sjkdhf] 0')  # tag preserved
```

---

## Status

Ready for review. `deltakit-circuit` tests pass (191, including the new regression
tests) and `ruff` / `mypy` / `pydoclint` are clean.

This overlaps the in-flight #248 (`stim` → `deltakit-stim`) migration. The
detection change is intentionally distribution-name-agnostic so it stays correct
before and after that lands — happy to coordinate or fold it in if you'd prefer.

---

## Future Work

Once `deltakit-stim`'s versioning relative to upstream `stim` is settled under
#248, the `_LOWEST_STIM_VERSION_WITH_TAG_FEATURE` threshold may want revisiting so
it maps onto the equivalent `deltakit-stim` version.

---

## Additional Information

- `tests/test_stim_parsing/test_tag_parsing.py`: the tag round-trip test previously
  skipped based on a raw `version("stim")` call, which itself raises when stim is
  installed under a different distribution name. It now gates on
  `is_stim_tag_feature_available()`.
- `tests/test_stim_parsing/test_stim_version_compatibility.py` (new): regression
  tests for the metadata lookup, the `__version__` fallback, the `None` path, and
  the availability threshold.

---

## Release Note

```
fix(circuit): detect stim version robustly so as_stim_circuit keeps tags (#262)
```

(If rebase-merged instead, the branch is already three clean Conventional-Commits
commits: one `fix` + two `test`.)

### DIFF
--- a/deltakit-circuit/tests/test_stim_parsing/test_tag_parsing.py
+++ b/deltakit-circuit/tests/test_stim_parsing/test_tag_parsing.py
@@ -15,7 +15,6 @@ from deltakit_circuit.gates._two_qubit_gates import TWO_QUBIT_GATES
 from deltakit_circuit.noise_channels._correlated_noise import ALL_CORRELATED_NOISE
 
 
-@pytest.mark.skip(reason="Mark as skipped until #262 resolution.")
 @pytest.mark.parametrize(
     ("instr_template", "tag"),
     list(


deltakit-circuit/src/deltakit_circuit/_stim_version_compatibility.py
```diff
@@ -1,9 +1,19 @@
 from importlib.metadata import PackageNotFoundError, version
 
-from packaging.version import Version
+from packaging.version import InvalidVersion, Version
```
rolandriver: Is this used somewhere?

deltakit-circuit/tests/test_stim_parsing/test_stim_version_compatibility.py
```diff
@@ -0,0 +1,54 @@
...
+If the installed stim version cannot be determined, ``is_stim_tag_feature_available``
+returns ``False`` and gate tags are silently dropped by ``Circuit.as_stim_circuit``.
+"""
+
+import sys
+import types
+from importlib.metadata import PackageNotFoundError
+
+import pytest
+import stim
```
rolandriver: The intention in #277 is to entirely drop importing `stim` in favor of `deltakit_stim`. So I don't think we should test against it.

deltakit-circuit/tests/test_stim_parsing/test_tag_parsing.py
rolandriver: This marker should be dropped to enable this test to fix the bug.

rolandriver (commented):
Thank you @lukedev45 for your interest in contributing to Deltakit, this is much appreciated. For this fix to work and since #277 is now merged, I believe all imports to stim must be dropped. I used `import deltakit_stim as stim` mainly for backwards compatibility but this must be rectified in the future.

lukedev45:
Thanks @rolandriver, I've taken another look and #277 already does fixes the root cause, so I've rebased onto current main and slimmed this down to just the regression test.

I dropped my _stim_version_compatibility.py rewrite
#277's version("deltakit_stim") lookup is the right approach, so there's nothing left to change there (and the InvalidVersion fallback you flagged is gone with it).
Removed test_stim_version_compatibility.py
Agreed it shouldn't import/test against stim.
Dropped the @pytest.mark.skip(…until #262 resolution) on test_parse_tagged_instruction with deltakit-stim>=0.2 (#296), is_stim_tag_feature_available() is True and as_stim_circuit() preserves tags, so the round-trip now passes (184 parametrized cases). 
No import stim left in the diff.
Net diff is a single line removed. Happy to fold this into a #277 follow-up instead if you'd prefer.

rolandriver:
> Thanks @rolandriver, I've taken another look and #277 already does fixes the root cause, so I've rebased onto current main and slimmed this down to just the regression test.
> 
> I dropped my _stim_version_compatibility.py rewrite #277's version("deltakit_stim") lookup is the right approach, so there's nothing left to change there (and the InvalidVersion fallback you flagged is gone with it). Removed test_stim_version_compatibility.py Agreed it shouldn't import/test against stim. Dropped the @pytest.mark.skip(…until #262 resolution) on test_parse_tagged_instruction with deltakit-stim>=0.2 (#296), is_stim_tag_feature_available() is True and as_stim_circuit() preserves tags, so the round-trip now passes (184 parametrized cases). No import stim left in the diff. Net diff is a single line removed. Happy to fold this into a #277 follow-up instead if you'd prefer.

Thank you @lukedev45. What I can suggest for this PR is to entirely remove all `is_stim_tag_feature_available()` related functionality as `deltakit-stim` is now backed up by at least `stim>=1.15` where tag features are _de facto_ present. This is now obsolete and can safely be removed.

rolandriver (commented):
As discussed in the other comment, I am happy to approve when you remove all tag feature checks and functionality. Thank you again.

lukedev45:
No problem. I will look into after work 🙂

rolandriver:
Hi @lukedev45 I have addressed it in #302. Thank you for your interest and contribution.

lukedev45:
Thanks @rolandriver ! I will look out for other things to work on. I'm working through the deltakit textbook at the moment and using deltakit as I go along, if I find anything I'll be sure to make a PR

PR #309, opened 2026-07-11
Remove ldpc dependency in deltakit-explore

### LINKED ISSUE #264: feat: replace implementation of kernel in GF2 from ldpc to galois
### Contributor License Agreement

- [x] A Contributor Licence Agreement signature is required for any first PR contribution. Further information can be found in our [contributing guidelines](CONTRIBUTING.md).

### Feature Description

Replace the method to compute kernel of parity matrices from `ldpc` to `galois` in the computation of logicals [here](https://github.com/Deltakit/deltakit/blob/8384d5a7378b98d49be8e0249347fab166e88920/deltakit-explorer/src/deltakit_explorer/codes/_logicals.py#L286) and drop the `ldpc` dependency.

### Minimal Reproducible Example

_No response_

### Actual behaviour

_No response_

### Expected behaviour

_No response_

### Additional context

_No response_

jsimbadev:
## 🔗 Closed Issues

Link any issues closed or comments addressed by this PR.
"Closes #X" will close issue X automatically when the PR is merged.

"Closes #264 "

---

## 📝 Description

This PR replaces the ldpc.mod2 usage in CSS logical-operator construction with galois GF(2) linear algebra in deltakit_explorer.codes._logicals.

  Specifically, it replaces the previous ldpc-based:

  - nullspace computation
  - rank computation
  - pivot-row selection

with a galois-based implementation over GF(2).

The new pivot-row helper was validated against ldpc.mod2.pivot_rows() on small dense binary matrices, and the logical-construction tests were expanded to check GF(2) correctness invariants rather than a specific basis ordering.

  The added tests verify that the returned logical operators:

  - lie in the appropriate kernel
  - extend the stabilizer row space to a basis for the kernel
  - preserve the expected logical dimension
  - are invariant under redundant parity-check rows

---

## 🚦 Status

Ready for review.

One follow-up question is whether the input contract to this path should be tightened further. The current implementation assumes binary parity-check matrices; if non-binary floating inputs were ever passed here, coercion to integer arrays before galois.GF2(...)   could silently normalize invalid inputs instead of rejecting them.

---

## 🛠️ Future Work

Potential follow-ups:

- tighten input validation / typing for binary GF(2) matrices
- remove any now-obsolete comments or dead code around the previous implementation
- drop the ldpc dependency if no other runtime usages remain

---

## ➕️ Additional Information

Anything that doesn't fit into a category above.

---

## 🧾 Release Note

If this PR will be **squash-merged**, draft the final one-line commit message here.
Otherwise, describe your rebase strategy or how you plan to maintain clean commits.

replace ldpc GF(2) logical-operator linear algebra with galois


nelimee (dismissed):
I think this is correct, but I think it needs more eyes on the maths and the tests just in case.

jsimbadev:
Anything else I can do to move things along just let me know @nelimee ^^

jsimbadev:
- Build docs / check documentation
      - This one may be transient. I have not investigated it in detail yet. Will find out on next push.

  - static-checks
      - pip-audit is failing with:
```
    Found 1 known vulnerability in 1 package
    Name  Version ID              Fix Versions
    ----- ------- --------------- ------------
    click 8.1.8   PYSEC-2026-2132 8.3.3
```
Options:
    - handle this in a separate issue / PR with a click bump
    - or include the bump in this PR if we want the branch to go green here

  - Test Deltakit
      - The Python 3.12 matrix failure appears to come from setup code in deltakit-explorer/tests/unit/circuits/test_tableau_functions.py, not from the new GF(2) logicals change. (***python version may be red herring***)
      - The test currently queries importlib.metadata with version("stim"), but I do not think importlib.metadata can resolve that from the import alias alone.
      - In a fresh environment:
```python
    >>> from importlib.metadata import Distribution
    >>> list(Distribution.discover(name="stim"))
    []
    >>> list(Distribution.discover(name="deltakit_stim"))
    [<importlib.metadata.PathDistribution ...>]
    >>> import deltakit_stim as stim
    >>> dir(stim)
    ['Circuit', 'CircuitErrorLocation', 'CircuitErrorLocationStackFrame', 'CircuitInstruction', 'CircuitRepeatBlock', ....]
```
I think the right fix is to query the installed package name directly while continuing to use import deltakit_stim as stim in the test file.

deltakit-explorer/tests/unit/circuits/test_tableau_functions.py
```diff
@@ -102,7 +102,7 @@
 )
 from deltakit_explorer.qpu._native_gate_set import NativeGateSet
 
-CURRENT_STIM_VERSION = Version(version("stim"))
+CURRENT_STIM_VERSION = Version(version("deltakit_stim"))
 STIM_VERSION_V1_13_0 = Version("1.13.0")
```
rolandriver: Thank you for this. There might be some remaining bits and pieces about `stim` that slipped through my attention. It feels like there are some here. It might be beyond the scope of this PR but if you could flag them in an issue, that'd be great for me to fix them up later.

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
```diff
@@ -241,6 +241,27 @@ def get_logical_operators_from_css_parity_check_matrices(
...
+    Compute the pivot row indices of a dense binary check matrix over GF(2).
+
+    Args:
+        check: A dense binary check matrix represented as a ``galois.FieldArray``
+            over GF(2).
+
+    Returns:
+        The pivot row indices of ``check``.
+    """
+    # Pivot rows of ``check`` are the pivot columns of ``check.T``.
```
rolandriver:
```suggestion
    # Pivot rows of ``check`` are the non-zero pivot columns of ``check.T``.
```

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
```diff
@@ -241,6 +241,27 @@ def get_logical_operators_from_css_parity_check_matrices(
...
+    rr = check.T.row_reduce()
```
rolandriver: Maybe more meaningful names than `rr` or `nz`.

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
```diff
@@ -274,22 +295,18 @@ def css_code_compute_logicals(
         a tuple ``(lx, lz)`` representing the X and Z logicals.
     """
 
-    def compute_lz(
+    def compute_lz_galois(
         _hx: NDArray[np.floating], _hz: NDArray[np.floating]
     ) -> NDArray[np.floating]:
-        # lz logical operators
-        # lz\in ker{hx} AND \notin Im(Hz.T)
+        _hx_gf = galois.GF2(np.asarray(_hx, dtype=np.int_))
```
rolandriver, line 362: Do we check that `_hx` and `_hz` are binary matrices? Do we reject them if not?

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
```diff
@@ -274,22 +295,18 @@ def css_code_compute_logicals(
         a tuple ``(lx, lz)`` representing the X and Z logicals.
     """
 
-    def compute_lz(
+    def compute_lz_galois(
```
rolandriver: Not sure this function name change is meaningful. I'd rather add this in a docstring that is missing.

deltakit-explorer/tests/unit/circuits/test_tableau_functions.py
```diff
@@ -102,7 +102,7 @@
 )
 from deltakit_explorer.qpu._native_gate_set import NativeGateSet
 
-CURRENT_STIM_VERSION = Version(version("stim"))
+CURRENT_STIM_VERSION = Version(version("deltakit_stim"))
```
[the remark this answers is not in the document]
jsimbadev: @rolandriver I can action this, though I think the machinery under the hood normalizes hence passing local examples. However I'm not sure what acceptable resolutions are for flaky linkcheck and the pip audit.

deltakit-explorer/tests/unit/codes/test_logicals.py
```diff
@@ -32,3 +36,129 @@ def test_paulistring_to_operator(string):
 #     operators_ref = [tuple(stim.PauliString(i) for i in j) for j in operators]
 #     operators_res = get_str_logical_operators_from_tableau(stabilisers)
 #     operators_res == operators_ref
+
+
+def _binary_matrix(rows: list[list[int]], n_cols: int) -> np.ndarray:
+    if len(rows) == 0:
+        return np.zeros((0, n_cols), dtype=np.uint8)
```
rolandriver: Do I understand this as returning the trivial binary matrix?

deltakit-explorer/tests/unit/codes/test_logicals.py
```diff
@@ -32,3 +36,129 @@ def test_paulistring_to_operator(string):
 #     operators_ref = [tuple(stim.PauliString(i) for i in j) for j in operators]
 #     operators_res = get_str_logical_operators_from_tableau(stabilisers)
 #     operators_res == operators_ref
+
+
+def _binary_matrix(rows: list[list[int]], n_cols: int) -> np.ndarray:
```
rolandriver: I would use `NDArray` from `np.typing` to make it more idiomatic to the code base.

deltakit-explorer/tests/unit/codes/test_logicals.py
```diff
@@ -32,3 +36,129 @@ def test_paulistring_to_operator(string):
...
+        vector = np.zeros(num_cols, dtype=np.uint8)
+        for coefficient, row in zip(coefficients, matrix, strict=True):
+            if coefficient:
+                vector ^= row
+        vectors.add(tuple(int(value) for value in vector))
+
+    return vectors
+
+
+def _binary_null_space(matrix: np.ndarray) -> set[tuple[int, ...]]:
```
rolandriver: Although these helper functions are supposed to be "private" in the Python sense, they probably deserve docstrings to explain what they are supposed to do. Here I think you are explicitly finding the nullspace by brute forcing all possible combinations of binary vectors that return 0 mod 2 while productised with the matrix. Am I right?

deltakit-explorer/tests/unit/codes/test_logicals.py
```diff
@@ -32,3 +36,129 @@ def test_paulistring_to_operator(string):
...
+
+
+def _dimension(space: set[tuple[int, ...]]) -> int:
+    return int(np.log2(len(space)))
+
+
+def _assert_logical_basis(
+    logicals: np.ndarray, kernel_checks: np.ndarray, image_checks: np.ndarray
+) -> None:
+    assert logicals.ndim == 2
```
rolandriver: Maybe being more explicit about these asserts.

deltakit-explorer/tests/unit/codes/test_logicals.py
```diff
@@ -32,3 +36,129 @@ def test_paulistring_to_operator(string):
...
+    assert np.all(np.isin(logicals, [0, 1]))
+
+    null_space = _binary_null_space(kernel_checks)
+    image_space = _binary_row_space(image_checks)
+    generated_space = _binary_row_space(np.vstack([image_checks, logicals]))
+
+    # CSS logicals should extend the image space into a basis for the kernel.
+    assert generated_space == null_space
+
+    expected_num_logicals = _dimension(null_space) - _dimension(image_space)
```
rolandriver: Is this checking the rank-nullity theorem?

deltakit-explorer/tests/unit/codes/test_logicals.py
```diff
@@ -32,3 +36,129 @@ def test_paulistring_to_operator(string):
...
+    assert logicals.shape[0] == expected_num_logicals
+
+
+@pytest.mark.parametrize(
+    ("hx", "hz"),
+    [
+        (_binary_matrix([], 3), _binary_matrix([], 3)),
```
rolandriver: Not too sure if it makes sense to pass an empty list and a `ncols` to create a binary matrix.

rolandriver (commented):
Thank you @jsimbadev for you contribution this is really appreciated. Comments from my side mainly for the sake of understanding.

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
jsimbadev: Done. Coming in next commit

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
jsimbadev: You're right! Went for `row_reduced` and `non_zero_row`. Coming in next commit

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
jsimbadev: I wasn't sure about this because prior this change, `NDArray[np.floating]` was being used, and since I didn't go outside of this file effectively, the call sites and other places most likely use this `np.floating` type.


The constructor for `galois.GF2` throws on float type arrays, and integers out of the Field range i.e >1 in GF2

```
galois/_fields/_array.py", line 192, in _convert_to_element
    raise TypeError(f"Valid element-like values are integers and string, not {type(element)}.")
```

and

```
galois/_fields/_array.py", line 177, in _verify_array_values
    raise ValueError(f"{cls.name} arrays must have elements in `0 <= x < {cls.order}`, not {values}.")
ValueError: GF(2) arrays must have elements in `0 <= x < 2`, not [2].
```
So this code will fail loud if the matrices are not binary in the GF2 sense. Do we want to explicitly check and also fail loud ourselves or let this bubble up? @rolandriver

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
jsimbadev: Yup you're right! Adding doc string also.

deltakit-explorer/tests/unit/codes/test_logicals.py
jsimbadev: yeah you're correct. I'd conflated the api is a little ambiguous because n_cols is optional so I've not split into two more explicit functions.

deltakit-explorer/tests/unit/codes/test_logicals.py
jsimbadev: Done )

deltakit-explorer/tests/unit/codes/test_logicals.py
jsimbadev: Yes you are right. I thought it reasonable for small tests, avoiding coupling to actual linear algebra library/logic.

deltakit-explorer/tests/unit/codes/test_logicals.py
jsimbadev: You're right. I'll name these variables so asserts read more naturally.

deltakit-explorer/tests/unit/codes/test_logicals.py
jsimbadev: yup, I've addressed that by splitting those two behaviours. The function was doing too much.

deltakit-explorer/tests/unit/codes/test_logicals.py
jsimbadev: Not exactly...

Note - I renamed `image_space` to `quotient_space`, since that seemed closer to its role here.

In this test, `null_space` is all operators that satisfy one set of checks, and `quotient_space` is the row space generated by the checks from the other side, i.e. the directions we treat as redundant. 

`compute_lz` is intended to return independent rows that satisfy the first checks, but are not already generated by the second set, so the expected number of rows is `dim(null_space) - dim(quotient_space)`

jsimbadev:
@rolandriver Tried to action your comments. Ready for review again

rolandriver:
Thank you @jsimbadev. I will review anytime soon. In the meantime, the fix for the failing static checks is to constrain the range of these dependencies: 

```toml
"tomlkit<0.16,>=0.13",
"python-semantic-release >=10.6.1, <11",
```

For checking the links, this is a timeout issue so rerunning it until it works (or until we find a better fix really).

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
```diff
@@ -241,6 +241,27 @@ def get_logical_operators_from_css_parity_check_matrices(
     )
 
 
+def pivot_rows(check: galois.FieldArray) -> NDArray[np.int_]:
```
rolandriver:
```suggestion
def pivot_rows(parity_check_matrix: galois.FieldArray) -> NDArray[np.int_]:
```

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
rolandriver: We can let it surface, no need to double check. Thanks for the explanation.

deltakit-explorer/tests/unit/circuits/test_tableau_functions.py
rolandriver: Could you please raise an issue pointing at this?

deltakit-explorer/tests/unit/codes/test_logicals.py
```diff
@@ -32,3 +36,185 @@ def test_paulistring_to_operator(string):
...
+
+def _dimension(space: set[tuple[int, ...]]) -> int:
+    return int(np.log2(len(space)))
+
+
+def _assert_logical_basis(
+    logicals: np.ndarray,
+    null_space_defining_checks: np.ndarray,
+    quotient_space_defining_checks: np.ndarray,
+) -> None:
```
rolandriver: Docstring.

rolandriver (commented):
Thank you for this @jsimbadev. I still see `np.array` in type annotations and return types that need to be replaced by `npt.NDArray`. Otherwise looking fine.

deltakit-explorer/src/deltakit_explorer/codes/_logicals.py
jsimbadev: Done. Coming with next commit )

jsimbadev:
raised issue https://github.com/Deltakit/deltakit/issues/313

I can move the patch as a fix against that issue. And I can make the suggested pyproject.toml suggestions there against that bug then hopefully get this GF2 PR in cleanly after a rebase.. thoughts @rolandriver

deltakit-explorer/tests/unit/codes/test_logicals.py
jsimbadev: Done )

AndrewPattersonRL (commented):
I've had a look at the implementation - which looks good to me! Thanks for your work on this @jsimbadev

AndrewPattersonRL:
Actually, on the tests - there are a small number of examples. I see that this is because the function that re-builds logicals without using the tested code is quite expensive. Could I suggest hard-coding a few examples? You can source some codes from `deltakit-explorer/src/deltakit_explorer/codes/_planar_code/_rotated_planar_code.py` and `deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py` which implement `parity_check_matrices`.

rolandriver:
@jsimbadev Could you please rebase, address the comment about testing by @AndrewPattersonRL and resolve any discussions? Thank you so much 🙏

jsimbadev:
Thanks for looking over the code everyone.

@AndrewPattersonRL following your suggestion to use actual codes led to find a bug in my implementation. 

It seems that not just any logical basis will do, after reading the through some of the BBCode logic, and noticed [a constraint on structure of the logical basis](https://github.com/Deltakit/deltakit/blob/8c5d6a74ba0a923ae26fe0c9c0f3bae073255611/deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py#L760) 
Which isn't really enforced by the implementation I've written, now do I see how the prior impl was, and now even I have a couple questions.

- Peeking into ldpc.mod2, they seem to just do standard linear algebra, yet yielded the structure required by BBCode. Was this by luck of convention or did I miss something in the implementation?
  - If this was implicit by convention in a transient library, should this enforcement be made explicit and where?
- There are no unit tests that exercise the constructor for BBCode, had there been I the bug I'd written would have been caught earlier, I plan to add such a test, should this be a pattern going forward?

@rolandriver @nelimee Would appreciate any advice )

Note: For reproducing, just try to construct a BBCode object from the feature branch.

AndrewPattersonRL:
Thanks, I'm having a look at this now. To your first point, I see that the matrices returned by both _are_ different, I'll dig a bit more to see where the convention is enforced. 

To your second point, this is a discussion for @rolandriver, @nelimee. When we migrated from the internal `code_explorer` we mustn't have taken any of the integration tests along. `BivariateBicycleCode` was only tested in integration due to the resources required. How do you propose testing BBCode going forward? There are some integration tests ready to go in `code_explorer`.

jsimbadev:
@AndrewPattersonRL Thanks for looking into this, apologies to put an extra thing on your plate.

With regards to the testing... In the past where I've had to deal with large or expensive constructions, the approaches often considered were the following

1.  Put all expensive things in integration or higher level testing
2. Build tests around invariants. There is already a lot of that going on [here](https://github.com/Deltakit/deltakit/blob/8c5d6a74ba0a923ae26fe0c9c0f3bae073255611/deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py#L461) 
For the change in question
In this case something like 
```
class BivariateBicycleCode(CSSCode):
...
def _ensure_logical_basis_structure(...):
    # make sure logical basis like X = [f | 0], Z=[h', g']
```
or have a test that calls a lazy BBCode._calculate_logical_operators and checks the above.

3. If possible refactor objects to be lazy on construction, only holding data necessary data, pushing the cost to calls that actually require the expensive work when needed. This may require some more thought, however it seems that the CSSCode class already has a [`calculate_logical_operators` option](https://github.com/Deltakit/deltakit/blob/8c5d6a74ba0a923ae26fe0c9c0f3bae073255611/deltakit-explorer/src/deltakit_explorer/codes/_css/_css_code.py#L106), so maybe could be viable.
4. Optimize to high heavens (typically last resort)

@rolandriver @nelimee Look forward to your thoughts

Note: However The above dont address the correctness issue.

jsimbadev:
I've written the following script to basically systematically isolate the point of divergence.

I generate kernels via `ldpc.mod2.nullspace(_hx).todense()` and inject them the downstream logical computation.

This suggests that the kernel space basis convention difference between the two libraries GF2 and ldpc (mod2) is worth looking into. @AndrewPattersonRL 

Let me know if you've found anything. I'll continue as I'm going

The following can be ran on my feature branch with a 
> uv add ldpc

```python

from __future__ import annotations

from dataclasses import dataclass
from functools import reduce

import galois
import numpy as np
import numpy.typing as npt
from ldpc import mod2

def pivot_rows(parity_check_matrix: galois.FieldArray) -> npt.NDArray[np.int_]:
    """Return pivot rows using the feature-branch GF(2) pivot logic."""
    row_reduced = parity_check_matrix.T.row_reduce()
    pivots: list[int] = []
    for row in row_reduced:
        non_zero_row = np.flatnonzero(row != 0)
        if non_zero_row.size:
            pivots.append(int(non_zero_row[0]))
    return np.asarray(pivots, dtype=np.int_)


@dataclass(frozen=True)
class BBFixture:
    """Parameters defining one bivariate bicycle ``hx``/``hz`` fixture."""

    name: str
    param_l: int
    param_m: int
    m_a_powers: tuple[int, int, int]
    m_b_powers: tuple[int, int, int]

    @property
    def n(self) -> int:
        return 2 * self.param_l * self.param_m


FIXTURES = (
    BBFixture(
        name="small-n30-k8",
        param_l=3,
        param_m=5,
        m_a_powers=(1, 1, 4),
        m_b_powers=(0, 1, 2),
    ),
    BBFixture(
        name="paper-72-12-6",
        param_l=6,
        param_m=6,
        m_a_powers=(3, 1, 2),
        m_b_powers=(3, 1, 2),
    ),
    BBFixture(
        name="paper-90-8-10",
        param_l=15,
        param_m=3,
        m_a_powers=(9, 1, 2),
        m_b_powers=(0, 2, 7),
    ),
)


def bb_hx_hz(
    fixture: BBFixture,
) -> tuple[npt.NDArray[np.uint8], npt.NDArray[np.uint8]]:
    """Construct ``Hx = [A B]`` and ``Hz = [B.T A.T]`` for a BB fixture."""
    shift_l = np.roll(np.eye(fixture.param_l, dtype=np.int_), shift=1, axis=1)
    shift_m = np.roll(np.eye(fixture.param_m, dtype=np.int_), shift=1, axis=1)
    identity_l = np.eye(fixture.param_l, dtype=np.int_)
    identity_m = np.eye(fixture.param_m, dtype=np.int_)

    x = np.kron(shift_l, identity_m)
    y = np.kron(identity_l, shift_m)

    a = (
        reduce(
            np.add,
            (
                np.linalg.matrix_power(x, fixture.m_a_powers[0]),
                np.linalg.matrix_power(y, fixture.m_a_powers[1]),
                np.linalg.matrix_power(y, fixture.m_a_powers[2]),
            ),
        )
        % 2
    )
    b = (
        reduce(
            np.add,
            (
                np.linalg.matrix_power(y, fixture.m_b_powers[0]),
                np.linalg.matrix_power(x, fixture.m_b_powers[1]),
                np.linalg.matrix_power(x, fixture.m_b_powers[2]),
            ),
        )
        % 2
    )

    return (
        np.hstack((a, b)).astype(np.uint8),
        np.hstack((b.T, a.T)).astype(np.uint8),
    )


def full_mod2_path(
    _hx: npt.NDArray[np.uint8],
    _hz: npt.NDArray[np.uint8],
) -> dict[str, npt.NDArray[np.integer] | int]:
    """Run the original ``ldpc.mod2`` ``compute_lz(_hx, _hz)`` steps."""
    kernel = mod2.nullspace(_hx).todense()
    log_stack = np.vstack((_hz, kernel))
    rank = mod2.rank(_hz)
    all_pivots = mod2.pivot_rows(log_stack)
    pivots = all_pivots[rank:]

    return {
        "kernel": kernel,
        "log_stack": log_stack,
        "rank": rank,
        "all_pivots": all_pivots,
        "pivots": pivots,
        "logicals": log_stack[pivots],
    }


def feature_path_with_mod2_kernel(
    _hz: npt.NDArray[np.uint8],
    kernel_hx: npt.NDArray[np.uint8],
) -> dict[str, npt.NDArray[np.integer] | int]:
    """Run the feature-branch rank/pivot path using a supplied ``ker(_hx)``."""
    _hz_gf = galois.GF2(_hz)
    kernel_hx_gf = galois.GF2(np.asarray(kernel_hx, dtype=np.int_))
    log_stack = np.vstack((_hz_gf, kernel_hx_gf))
    rank = int(np.linalg.matrix_rank(_hz_gf))
    all_pivots = pivot_rows(log_stack)
    pivots = all_pivots[rank:]

    return {
        "log_stack": log_stack,
        "rank": rank,
        "all_pivots": all_pivots,
        "pivots": pivots,
        "logicals": log_stack[pivots],
    }


def right_zero_count(matrix: npt.NDArray[np.uint8]) -> int:
    """Return how many rows have zero support in the right block."""
    half = matrix.shape[1] // 2
    return int(np.sum(np.all(matrix[:, half:] == 0, axis=1)))


def compare_arrays(
    old_path: dict[str, npt.NDArray[np.integer] | int],
    feature_path: dict[str, npt.NDArray[np.integer] | int],
) -> dict[str, bool]:
    """Compare shared outputs from the two paths."""
    return {
        key: (
            old_path[key] == feature_path[key]
            if key == "rank"
            else bool(np.array_equal(old_path[key], feature_path[key]))
        )
        for key in ("log_stack", "rank", "all_pivots", "pivots", "logicals")
    }


def report_side(
    fixture: BBFixture,
    side_name: str,
    _hx: npt.NDArray[np.uint8],
    _hz: npt.NDArray[np.uint8],
) -> bool:
    """Print the comparison for one logical side."""
    old_path = full_mod2_path(_hx, _hz)
    feature_path = feature_path_with_mod2_kernel(
        _hz=_hz,
        kernel_hx=old_path["kernel"],
    )
    comparisons = compare_arrays(old_path, feature_path)
    matched = all(comparisons.values())

    print(f"  {side_name}:")
    print(f"    kernel shape: {old_path['kernel'].shape}")
    print(f"    log_stack shape: {old_path['log_stack'].shape}")
    print(f"    rank: mod2={old_path['rank']} feature={feature_path['rank']}")
    print(f"    mod2 pivots:   {old_path['pivots'].tolist()}")
    print(f"    feature pivots:{feature_path['pivots'].tolist()}")
    print(f"    exact matches: {comparisons}")
    print(
        "    right-zero logical rows: "
        f"{right_zero_count(old_path['logicals'])} / "
        f"{old_path['logicals'].shape[0]}"
    )

    if not matched:
        print(f"    FAILED: {fixture.name} {side_name} did not match exactly.")

    return matched


def report_fixture(fixture: BBFixture) -> bool:
    """Print the comparison for one BB fixture."""
    hx, hz = bb_hx_hz(fixture)
    rank_hx = int(mod2.rank(hx))
    rank_hz = int(mod2.rank(hz))
    logical_dimension = fixture.n - rank_hx - rank_hz

    print(f"\n{fixture.name}")
    print(
        "  parameters: "
        f"l={fixture.param_l}, m={fixture.param_m}, "
        f"A={list(fixture.m_a_powers)}, B={list(fixture.m_b_powers)}"
    )
    print(f"  shapes: hx={hx.shape}, hz={hz.shape}")
    print(f"  ranks: rank(hx)={rank_hx}, rank(hz)={rank_hz}, k={logical_dimension}")
    print(f"  checks commute: {bool(np.all((hx @ hz.T) % 2 == 0))}")

    # Match css_code_compute_logicals: lx = compute_lz(hz, hx), lz = compute_lz(hx, hz).
    lx_matched = report_side(fixture, "lx = compute_lz(hz, hx)", _hx=hz, _hz=hx)
    lz_matched = report_side(fixture, "lz = compute_lz(hx, hz)", _hx=hx, _hz=hz)

    return lx_matched and lz_matched


def main() -> None:
    """Run the diagnostic comparison for all fixtures."""
    #print(f"pivot_rows source: {PIVOT_ROWS_SOURCE}")
    all_matched = True
    for fixture in FIXTURES:
        all_matched = report_fixture(fixture) and all_matched

    print("\nSummary")
    print(f"  all fixtures matched exactly: {all_matched}")

    if not all_matched:
        raise SystemExit(1)


if __name__ == "__main__":
    main()
```

AndrewPattersonRL:
Yeah, I've been looking into it also, there's some AI-gen code that computes preferred basis vectors for the BB code - and then requests these from your get_logicals function. I'll add it here for review, when I've had a little longer to look at it.

AndrewPattersonRL:
Hi, there's some proposed changes to the `get_logicals` function, where is accepts preferred basis vectors:

```
def css_code_compute_logicals(
    hx: NDArray[np.floating],
    hz: NDArray[np.floating],
    *,
    lx_preferred: NDArray[np.floating] | None = None,
    lz_preferred: NDArray[np.floating] | None = None,
) -> tuple[NDArray[np.floating], NDArray[np.floating]]:
    """Drop-in replacement for calling bposd.css_code.compute_logicals.

    Note:
        This function has been taken from the following repository under the MIT
        licence: [bp_osd repository](https://github.com/quantumgizmos/bp_osd).

        The following modifications were performed:
            1. Add more detailed typing information (parameter types, return type).
            2. Remove two lines of code that were not directly used by the method:
                ```
                if self.K == np.nan:
                    self.compute_dimension()
                ```
            3. Change the names of the parameters of compute_lz to avoid a name clash.
            4. Remove all the `self.`.
            5. Add a docstring.
            6. Add typing to the internal `compute_lz` function.
            7. Only use dense matrices because the inputs are dense anyway.
            8. Add ``lx_preferred``/``lz_preferred`` to let a caller bias which
               representative is picked for a logical class, since the basis a
               plain kernel/quotient computation returns is not unique and a
               caller may rely on a structural property (e.g. sparsity, or a
               particular support pattern) that isn't guaranteed by an
               arbitrary basis.

        You can check the original version of this function at
        [this permalink](https://github.com/quantumgizmos/bp_osd/blob/8894ec654b24ae875c07e5a361dcae9a77d748ce/src/bposd/css.py#L75).

    Args:
        hx: parity check matrix for the X code.
        hz: parity check matrix for the Z code.
        lx_preferred: optional binary vectors, each already lying in
            ``ker(hz)``, to prefer as representatives of their logical class.
            Useful when a caller needs representatives with a specific
            structural property that a generic nullspace basis wouldn't
            otherwise guarantee. Vectors that are stabilisers or dependent on
            earlier rows are skipped and contribute nothing extra.
        lz_preferred: as ``lx_preferred``, but for Z logicals, each already
            lying in ``ker(hx)``.

    Returns:
        a tuple ``(lx, lz)`` representing the X and Z logicals.
    """

    def compute_lz(
        _hx: NDArray[np.floating],
        _hz: NDArray[np.floating],
        _preferred: NDArray[np.floating] | None,
    ) -> NDArray[np.floating]:
        """Compute a basis of logical operators for one side of a CSS code.

        This function finds operators that satisfy the checks in ``_hx`` but are
        not just combinations of the checks in ``_hz``.

        Args:
            _hx: Check matrix used to define valid operators.
            _hz: Check matrix used to remove redundant operators.
            _preferred: Optional vectors, already lying in ``ker(_hx)``, to
                prefer as representatives of their logical class over other
                representatives that would otherwise be found.

        Returns:
            A binary matrix whose rows are logical operators.
        """
        _hx_gf = galois.GF2(np.asarray(_hx, dtype=np.int_))
        _hz_gf = galois.GF2(np.asarray(_hz, dtype=np.int_))

        ker_hx_gf = _hx_gf.null_space()
        rank_hz_gf = np.linalg.matrix_rank(_hz_gf)

        stack = [_hz_gf]
        if _preferred is not None:
            stack.append(galois.GF2(np.asarray(_preferred, dtype=np.int_) % 2))
        stack.append(ker_hx_gf)
        log_stack = np.vstack(stack)
        pivots = pivot_rows(log_stack)[rank_hz_gf:]

        return np.asarray(log_stack[pivots])

    return (
        compute_lz(hz, hx, lx_preferred),
        compute_lz(hx, hz, lz_preferred),
    )
```

And in the BB code get logicals function we also do:

```
    def _get_logicals(
        self,
    ) -> tuple[list[list[PauliX[Coord2D]]], list[list[PauliZ[Coord2D]]]]:
        """
        Will return a tuple of lists of the X and Z logicals, respectively.
        They are paired up in order, that is, they anticommute when the
        indices of the two lists are aligned and commute otherwise.

        Returns
        -------
        Tuple[List[List[PauliX[Coord2D]]], List[List[PauliZ[Coord2D]]]]
            Tuple of X and Z logical operators, in order such that
            x_logicals[i],z_logicals[j] anti-commute exactly when i=j
            and commute otherwise.
        """
        # get a subset of logicals from css_code, and compute the rest.
        #
        # X logicals are of the form X(f,0) and X(g,h), eq. (16). A basis of
        # ker(hz) mod im(hx.T) does not necessarily contain any X(f,0)-type
        # representatives (the split is a property of a *class*, not of an
        # arbitrary representative chosen for it), so pure f-type kernel
        # vectors -- i.e. ker(hz[:, :half]) embedded as (f, 0) -- are passed
        # as preferred representatives, guaranteeing as many X(f,0) logicals
        # as exist are found before falling back to mixed X(g,h) logicals.
        half = self.n // 2
        hz_gf = galois.GF2(np.asarray(self.m_Hz, dtype=np.int_))
        pure_f_kernel = hz_gf[:, :half].null_space()
        pure_f_logs = np.hstack(
            (
                np.asarray(pure_f_kernel, dtype=np.int_),
                np.zeros((pure_f_kernel.shape[0], half), dtype=np.int_),
            )
        )

        x_logs, _ = css_code_compute_logicals(
            self.m_Hx, self.m_Hz, lx_preferred=pure_f_logs
        )

[... the rest is the same]
```

I've validated this against the codes in the original IBM paper. What do you think of the suggested changes?

jsimbadev:
Morning @AndrewPattersonRL, thanks for getting back with your suggestion. I think this is a neat approach. I had not thought about getting `(f | 0)` candidates directly from `ker(Hz[:, :half])` and then reusing
  `css_code_compute_logicals`.

  I think there are a few details that need a little care:

  - validation of the preferred rows, to ensure they really lie in the relevant kernel i.e `_preferred @ _hi.T == 0` for each branch (`compute_lz`, `compute_lx`).
  - I do not think we should apply `% 2` when constructing `_preferred` as a `galois.GF2` object; that would silently accept non-binary input. Since we want fail-loud semantics, we should let `galois.GF2(...)` reject invalid values.
  - If I understand the `stack` correctly, there is a hidden dependence on the convention of `pivot_rows` and the order in which it works. I will rename pivot_rows to something like `independent_row_indices_in_order`. And of course document.

  Should we also add a way to compute only one side? In the BB path we only use `x_logs`, so with the current API we would do the extra `lz` work only to discard it.
  
  Of course I'll add some tests with proper BB codes as @rolandriver suggested earlier. Sound like a plan?

rolandriver:
Thank you for this @jsimbadev really. Some questions/comments from my side:

> Morning @AndrewPattersonRL, thanks for getting back with your suggestion. I think this is a neat approach. I had not thought about getting `(f | 0)` candidates directly from `ker(Hz[:, :half])` and then reusing `css_code_compute_logicals`.
> 
> I think there are a few details that need a little care:
> 
> * validation of the preferred rows, to ensure they really lie in the relevant kernel i.e `_preferred @ _hi.T == 0` for each branch (`compute_lz`, `compute_lx`).

I think being defensive here is a good idea.

> * I do not think we should apply `% 2` when constructing `_preferred` as a `galois.GF2` object; that would silently accept non-binary input. Since we want fail-loud semantics, we should let `galois.GF2(...)` reject invalid values.

Can this be added to the validation check above? My understanding here is that this is an implicit check for `preferred` to be a binary matrix. I would suggest to group them into a check that:
  - Checks `preferred` are genuinely binary
  - Really belong to the kernel of the parity-check matrix.

> * If I understand the `stack` correctly, there is a hidden dependence on the convention of `pivot_rows` and the order in which it works. I will rename pivot_rows to something like `independent_row_indices_in_order`. And of course document.
> 
> Should we also add a way to compute only one side? In the BB path we only use `x_logs`, so with the current API we would do the extra `lz` work only to discard it.
> 
> Of course I'll add some tests with proper BB codes as @rolandriver suggested earlier. Sound like a plan?

Definitely sounds like a plan ! Thanks again.

rolandriver:
Hey @jsimbadev any further thoughts on this? I am keen on moving on this if possible.

jsimbadev:
@rolandriver yeah I'm part way through the above plan... Will submit tonight or tomorrow!

rolandriver:
> @rolandriver yeah I'm part way through the above plan... Will submit tonight or tomorrow!

Super nice, thank you so much 🙏

jsimbadev:
I’ve reworked this so the tests are checking the intended contracts rather than comparing against a particular logical basis.

The main change is that the generic CSS logical computation now supports preferred representatives. BBCode uses this to bias the X logical basis toward the BB-structured representatives it needs, instead of relying on whatever arbitrary valid CSS basis happens to be returned.

I also split the BB construction up a bit. Parity-check matrix construction is now factored out of the constructor so tests can use `Hx`/`Hz` directly without constructing the full code object. The BB logical-vector construction is also exposed separately as `compute_bb_structured_logicals(...)`, and the constructor just maps those vectors onto the code’s qubit layout afterwards.

The tests are split by responsibility:

 - `test_logicals.py` checks the generic CSS contract: the returned logicals are valid bases for representative code fixtures, including BB and planar codes.
 -  - `test_bivariate_bicycle_code.py` checks BB-specific behavior only: the parity-check matrices have the expected dimensions/commutation properties, and the BB-structured logical construction preserves the pure `[f | 0]` logical classes modulo X stabilisers.

This avoids hard-coding exact logical vectors or row ordering, which would be brittle because the basis choice is not unique, while still checking the BB-specific structure.

@rolandriver @AndrewPattersonRL

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
```diff
@@ -23,6 +24,26 @@
 from deltakit_explorer.codes._stabiliser import Stabiliser
 
 
+class _BivariateBicycleMatrices(NamedTuple):
+    """Matrices used to construct a bivariate bicycle code.
+
+    Attributes:
+        m_x: Binary matrix representing ``x`` in the IBM paper.
```
rolandriver: What IBM paper ? Could you please cite it in the docstring?

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
```diff
@@ -23,6 +24,26 @@
...
+        m_y: Binary matrix representing ``y`` in the IBM paper.
+        m_A_submatrices: Submatrices whose sum defines matrix ``A``.
+        m_B_submatrices: Submatrices whose sum defines matrix ``B``.
```
rolandriver: I assume the `m` prefix means "matrix". Not too sure this is totally useful as I read this one as "matrix A submatrices" which sounds redundant. Since it is explicit from the description these are matrices, I'd remove it.

rolandriver: On that note, could you please elaborate on what `x`, `y`, `A` and `B` are ?

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
```diff
@@ -387,45 +409,20 @@ def __init__(
...
-            # assert properties of A and B:
-            self._assert_matrix_ma_mb_properties(m_A, m_B)
-        self.m_A_submatrices = (m_A1, m_A2, m_A3)
-        self.m_B_submatrices = (m_B1, m_B2, m_B3)
-
-        # create Hx and Hz check matrices. Note that since A and B
-        # commute, these are valid check matrices
-        self.m_Hx = np.hstack((m_A, m_B))
-        self.m_Hz = np.hstack((m_B.T, m_A.T))
+        self.m_x = matrices.m_x
```
rolandriver: It is even more obvious here as it reads "from matrices get me matrix x" and it would be cleaner to read "from matrices get me x".

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
```diff
@@ -457,6 +468,63 @@ def __init__(
...
+    def _construct_matrices(
+        param_l: int,
+        param_m: int,
+        m_A_powers: list[int],
+        m_B_powers: list[int],
+    ) -> _BivariateBicycleMatrices:
+        """Construct the matrices used by a bivariate bicycle code.
+
+        Args:
+            param_l: Parameter ``l`` as in the IBM paper.
```
rolandriver: Reference to paper.

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
```diff
@@ -457,6 +468,63 @@ def __init__(
...
+            param_m: Parameter ``m`` as in the IBM paper.
+            m_A_powers: Powers defining matrix ``A``.
+            m_B_powers: Powers defining matrix ``B``.
+
+        Returns:
+            The shift, polynomial submatrices, and parity check matrices used
```
rolandriver: More explanation what these are.

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
```diff
@@ -457,6 +468,63 @@ def __init__(
             check_logical_operators_are_independent=check_logical_operators_are_independent,
         )
 
+    @staticmethod
+    def _construct_matrices(
```
rolandriver:
```suggestion
    def _construct_bb_matrices(
```

Or smtg around this.

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
```diff
@@ -739,23 +807,67 @@ def _get_stabilisers(self) -> list[Stabiliser]:
...
-        indices of the two lists are aligned and commute otherwise.
+        m_Hx = np.asarray(m_Hx, dtype=np.int_)
+        m_Hz = np.asarray(m_Hz, dtype=np.int_)
+        expected_shape = (param_l * param_m, 2 * param_l * param_m)
+        if m_Hx.shape != expected_shape or m_Hz.shape != expected_shape:
+            msg = (
+                "BB parity check matrices should both have shape "
+                f"{expected_shape}, got {m_Hx.shape} and {m_Hz.shape}."
+            )
+            raise ValueError(msg)
```
rolandriver, line 881: Is it really a `ValueError` here?

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
```diff
@@ -739,23 +807,67 @@ def _get_stabilisers(self) -> list[Stabiliser]:
...
-        Tuple[List[List[PauliX[Coord2D]]], List[List[PauliZ[Coord2D]]]]
-            Tuple of X and Z logical operators, in order such that
-            x_logicals[i],z_logicals[j] anti-commute exactly when i=j
-            and commute otherwise.
-        """
-        # get a subset of logicals from css_code, and compute the rest
-        x_logs, _ = css_code_compute_logicals(self.m_Hx, self.m_Hz)
+        # get a subset of logicals from css_code, and compute the rest.
+        #
+        # X logicals are of the form X(f,0) and X(g,h), eq. (16). A basis of
```
rolandriver, line 885: Not too sure what these forms are. Can you please elaborate?

rolandriver (dismissed):
Looking good for me. Approving now but subjected to @AndrewPattersonRL 's validation.

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
jsimbadev: https://arxiv.org/pdf/2308.07915, adding in doc strings where necessary

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
jsimbadev: You're right in this context it is redundant, I'll also add some words about what these matrices are as per the paper.
The BB construction is based on coordinate wise cyclic shifts, (coords `l` and `m` arbitrarily) ...
```
x = S_l \tensorI_m, y = I_m \tensor S_m
``` 
Then `A` and `B` are constructed as 

```
A = A1 + A2 + A3, B = B1 + B2 + B3
```

Each `A_i` and `B_i` are expressed as a power of `x and y`... hence they are parameterized via 2 integers in the code `m_A_powers: list[int]` and `m_B_powers: list[int]` 
... 

Putting something to this effect now

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
jsimbadev: Yup, agree on this

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
jsimbadev: Personally I think so, because the inputs are valid array-like arguments, but their shapes are invalid for the BB parameters supplied.

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
jsimbadev: These forms are the BB-specific logical representative families from Eq. (16) of Bravyi et al. In the BB construction the data-qubit vector is split into left/right halves, so an X logical vector can be written as `(left, right)`.

The paper chooses polynomials/vectors `f`, `g`, and `h` satisfying the BB kernel constraints, then uses them to build two symmetric logical blocks:

  `X(alpha f, 0)`, `Z(alpha h.T, alpha g.T)`
  `X(alpha g, alpha h)`, `Z(0, alpha f.T)`

  for monomial shifts `alpha`.

This part is not exactly my expertise; I was working at the level of linear systems and constraints. My understanding is that an arbitrary CSS logical basis need not expose representatives with the pure right-zero shape `(f, 0)`. That is why this path biases `css_code_compute_logicals` with the available pure-`f` representatives.

Section 9.4 of the paper explains why this structure is useful.

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
jsimbadev: I enriched the docstring, let me know if you think its still lacking in clarity

jsimbadev:
I think something is up with the CI because pre-commit is doing just fine. Perhaps a mismatch between ruff versions locally and remotely ... I'll look into this a little further in a little bit

deltakit-explorer/src/deltakit_explorer/codes/_bivariate_bicycle_code.py
rolandriver: Thank you. If that could be notified in the docstring that'd be great.

jsimbadev:
@rolandriver this is adding a lot of noise to the PR, and I suspect incoming PRs too, I think it ought to be pinned and a separate issue raised to either address the style errors.

$ uvx ruff@0.16.1 check deltakit-decode/src/deltakit_decode/analysis/_decoding_circuits.py
Installed 1 package in 4ms
PLR0917 Too many positional arguments (6 > 5)
  --> deltakit-decode/src/deltakit_decode/analysis/_decoding_circuits.py:10:5
   |
10 | def run_decoding_on_circuit(
   |     ^^^^^^^^^^^^^^^^^^^^^^^
11 |     circuit: stim.Circuit,
12 |     max_shots: int,
   |

Found 1 error.
$ uvx ruff@0.15.21 check deltakit-decode/src/deltakit_decode/analysis/_decoding_circuits.py
Installed 1 package in 9ms
All checks passed!

jsimbadev:
@rolandriver `"ruff>=0.14,<0.16"`?

rolandriver:
> @rolandriver `"ruff>=0.14,<0.16"`?

You can get away with it as I did [here](https://github.com/Deltakit/deltakit-decode/blob/24515557b2a18822006726bba27b25f9d8b69b9b/ruff.toml#L109).

deltakit-explorer/tests/unit/codes/test_logicals.py
```diff
@@ -32,3 +47,240 @@ def test_paulistring_to_operator(string):
...
+        ),
+        (
+            _binary_matrix([[1, 1, 1, 1]]),
+            _binary_matrix([[1, 1, 0, 0], [0, 0, 1, 1]]),
+        ),
+    ],
+)
+def test_css_code_compute_logicals_returns_valid_css_logical_bases(
+    hx: npt.NDArray[np.uint8], hz: npt.NDArray[np.uint8]
+):
```
rolandriver:
```suggestion
) -> None:
```

deltakit-explorer/tests/unit/codes/test_logicals.py
```diff
@@ -32,3 +47,240 @@ def test_paulistring_to_operator(string):
...
+    assert_valid_css_logical_basis(
+        lz,
+        hx,
+        hz,
+        null_space_dimension=_dimension(_binary_null_space(hx)),
+        quotient_space_dimension=_dimension(_binary_row_space(hz)),
+    )
+
+
+def test_css_code_compute_logicals_is_invariant_to_redundant_rows():
```
rolandriver: Same here and for all test where it is missing.

rolandriver (dismissed):
Excellent work, thank you so much @jsimbadev. Looking good to me.

rolandriver (approved):
Re-approving as it has been dismissed. @jsimbadev feel free to resolve and close discussions whenever appropriate.

jsimbadev:
Thanks for the guidance and patience through the process @rolandriver @AndrewPattersonRL @nelimee Glad I was able to contribute )

AndrewPattersonRL:
Thanks very much @jsimbadev!

Issue #325, opened 2026-08-02
request: sliding-window and parallel-window decoding on top of the existing deltakit-core window primitives


zkasuran:
## Summary

`deltakit-core` publicly exports twelve window-surgery primitives across `_windowing_utils.py` and `_hypergraph_annotations_tools.py` (imported in `decoding_graphs/__init__.py` and listed in `__all__`). Nothing in Deltakit consumes them. They arrived in #160. A windowed (sliding-window) decoder is the natural consumer: decode a long experiment in overlapping time slices, commit the middle of each slice, hand the committed corrections that dangle out of the commit region to the next window as artificial defects. That is the overlapping-recovery method from arXiv:2209.08552.

I probed the primitives on a `RotatedPlanarCode(5, 5)` memory circuit and they compose into a per-window graph cleanly, so the gap is the decoder-level algorithm (commit and buffer split, artificial-defect handoff, cross-window logical accounting, the parallel schedule), not the graph handling. The probe output and the full design are in the fold below.

I would build it in `deltakit-decode` implementing the existing `DecoderProtocol`, wrapping any inner decoder that returns a full correction so `PyMatchingDecoder` works unchanged. Sequential sliding window first, the parallel two-layer schedule from the same paper as a follow-up.

**Status.** @rolandriver has deferred this on the roadmap to early next year, so it is parked rather than active. The one self-contained piece, `weighted_nodes_within_radius` (the last `NotImplementedError` in `decoding_graphs`), is already done and under review as Deltakit/deltakit-core#12. This issue has been condensed at @nelimee's request; the original text is preserved unchanged in the fold.

<details>
<summary>Full original proposal (primitives, probe numbers, design, acceptance criteria, alternatives, open questions, environment)</summary>

### Contributor License Agreement

- [x] A Contributor License Agreement signature is requested for any first PR contribution. Further information can be found in our [contributing guidelines](CONTRIBUTING.md).

### Motivation

`deltakit-core` exports a complete set of window-surgery primitives and nothing in Deltakit uses them.

Checked on `main` at `8c5d6a7`. `deltakit_core.decoding_graphs` publicly exports twelve functions across two modules:

`_windowing_utils.py`
- `nodes_within_radius` (L16)
- `weighted_nodes_within_radius` (L54)
- `induce_subhypergraph` (L77)
- `connect_dangling_to_boundary_hypergraph` (L105)
- `relabel_hypergraph_nodes_contiguously` (L159)
[...]

PR #326, opened 2026-08-03
feat(core): implement weighted_nodes_within_radius

zkasuran:
## 🔗 Closed Issues

Nothing closes with this. It is one of the small independent pieces #325 offers to do
on its own, and the one the sliding window in that proposal needs first.

---

## 📝 Description

`weighted_nodes_within_radius` is in `deltakit_core.decoding_graphs.__all__` and raises
`NotImplementedError`. It is the only `NotImplementedError` in `decoding_graphs`. Its
docstring already specifies the answer, "nodes reachable with total path weight
<= radius", so this implements exactly that: Dijkstra with a budget, sitting next to
`nodes_within_radius`, which counts hops.

Cost accumulates per edge traversed, so every vertex of a hyperedge is the same distance
from the vertex being left.

`EdgeRecord.weight` is `log((1 - p_err) / p_err)`. Two consequences of that belong in the
docstring rather than being left for a caller to discover, and both are tested.

- **`p_err` of 0, the `EdgeRecord` default, gives an infinite weight.** An error mechanism
  that never fires is infinitely far away and growth stops at it, so no budget crosses a
  default edge. That is the intended reading and it is why this is not a drop-in
  replacement for the hop counter. One test asserts the two functions differ on the same
  graph for that reason.
- **`p_err` above 0.5 gives a negative weight.** There is no well defined shortest path
  then, and the returned set would depend on visit order, so it raises `ValueError`
  naming the edge instead of returning something order-dependent.

A negative radius raises `ValueError` with the same message `nodes_within_radius` already
uses, and `radius == 0` returns the start nodes, so the two functions agree at their edges.

Seven tests in a new `TestWeightedNodesWithinRadius`:

| Test | What it pins |
| --- | --- |
| `test_budget_is_spent_per_edge_not_per_hop` | the budget buys edge weight, walked up in steps of `log(9)` |
| `test_every_vertex_of_a_hyperedge_costs_the_same_to_reach` | one hyperedge, one cost, for all of its vertices |
| `test_a_cheaper_route_wins_over_a_shorter_one` | two cheap edges beat one dear edge, so this is not BFS with weights bolted on |
| `test_an_impossible_error_mechanism_is_infinitely_far` | a default `EdgeRecord` stops growth where the hop version walks through |
| `test_several_start_nodes_grow_together` | multi-source, seeded at distance 0 |
| `test_negative_radius_is_refused` | matches the hop version's message |
| `test_negative_edge_weight_is_refused_rather_than_guessed` | `p_err = 0.9` raises rather than returning an order-dependent set |

Verified locally on this branch:

```
19606 passed, 6 skipped in 433.21s        # full suite, uv run pytest
                                          # 19599 passed, 6 skipped on 8c5d6a7
deltakit-core static checks, all clean:
  typos .                    (silent)
  ruff check .               All checks passed!
  ruff format --check .      36 files already formatted
  pydoclint .                🎉 No violations 🎉
  mypy .                     Success: no issues found in 36 source files
```

---

## 🚦 Status

Ready to review.

Correction, 2026-08-04: this section used to say `static-checks` was red on `main` over
`PLR0917` and that the other packages in the matrix would fail until #324 landed. That is
no longer true. 50ca5a3 added `PLR0917` to the package ruff configs, the matrix is green
with `ruff==0.16.1`, and #324 is closed as unnecessary. Nothing in this change ever
depended on it.

---

## 🛠️ Future Work

The sequential and parallel sliding-window decoders in #325, which use this together with
`annotate_edges_with_window_ids` and `get_window_id_transfer_graph`. Nothing else in
`_windowing_utils.py` is unimplemented after this.

---

## ➕️ Additional Information

Design decision worth flagging, because it is a judgement call rather than a
transcription of the docstring: an inclusive `<= radius` boundary is compared against a
float that was accumulated by addition. The tests therefore build their expected budgets
by the same arithmetic the function performs (`step`, `2 * step`) instead of hardcoding
decimals, so they pin behaviour rather than a rounding.

AI assistance (Claude, Anthropic) was used in developing this change. The design, review
and verification were done by the author. Verified locally before submitting: the full
`pytest` suite, plus `typos`, `ruff check`, `ruff format --check`, `pydoclint` and `mypy`
on `deltakit-core`, with the output quoted above.

---

## 🧾 Release Note

`feat(core): implement weighted_nodes_within_radius as bounded Dijkstra over edge weights`

### DIFF
--- a/deltakit-core/src/deltakit_core/decoding_graphs/_windowing_utils.py
+++ b/deltakit-core/src/deltakit_core/decoding_graphs/_windowing_utils.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 
 from collections import deque
 from collections.abc import Iterable
+from heapq import heapify, heappop, heappush
+from math import inf
 
 from deltakit_core.decoding_graphs._data_qubits import EdgeRecord
 from deltakit_core.decoding_graphs._decoding_graph import (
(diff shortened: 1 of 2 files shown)


zkasuran:
Small update plus one request.

The note in the description about `static-checks` being red on `main` is out of date, so I have corrected it in place. 50ca5a3 ignores `PLR0917` in the package ruff configs and the matrix is green with `ruff==0.16.1`, so #324 is closed as unnecessary. Nothing in this change depended on it.

The request: this comes from a fork, so the test and static-check jobs are still waiting on a workflow approval and only `label-prs` and the CLA check have reported. Whenever suits you. It was green locally when I opened it, `19606 passed, 6 skipped` against `19599 passed, 6 skipped` on the base, with `deltakit-core`'s five static checks clean.

rolandriver:
@zkasuran Thank you for you contribution really. However we can't proceed with it now as subpackages in `deltakit` are expected to be removed imminently in favour of standalone repos. As such, could you please reopen your PR in the dedicated [deltakit-core repo](https://github.com/Deltakit/deltakit-core)? Thank you very much.

zkasuran:
Moved, thanks for the pointer. Deltakit/deltakit-core#12 is the same change in the
standalone repo.

Nothing moved but the paths, `deltakit-core/src/...` to `src/...` and the tests likewise.
The patch applied clean with no other edit, because the two copies of
`_windowing_utils.py` differ only in the copyright header. Re-verified over there against
that repo's own suite: 896 passed, 3 skipped, against 889 passed, 3 skipped on its `main`
at 01eecea, plus the eight static checks its Pull Request workflow runs (`typos`,
`ruff check`, `ruff format --check`, `pydoclint`, `mypy`, `deptry`, `bandit`,
`pip-audit`), all green.

It comes from a fork there too, so it needs the same workflow approval whenever suits you.

Closing this one. #332 made the layout obvious.

PR #332, opened 2026-08-06
feat: Remove subpackages

rolandriver:
## 🔗 Closed Issues

Closes #336.

---

## 📝 Description

Remove packages that are now in their own repos.

---

## 🚦 Status

Ready to review.

---

## 🛠️ Future Work

Focused PRs are good!
Create and link issues for any follow-up work needed.

---

## ➕️ Additional Information

Anything that doesn't fit into a category above.

---

## 🧾 Release Note

If this PR will be **squash-merged**, draft the final one-line commit message here.
Otherwise, describe your rebase strategy or how you plan to maintain clean commits.

Issue #45, opened 2025-09-19
request: add deltakit_decode integration tests


nelimee:
### Motivation

Integration tests for `deltakit_decode` exist, but are not included in the repository.

### Proposed solution

Add the integration tests.

mrviind:
I've opened #335 with three newly written, CI-fast integration tests toward this.
I'm aware of the `after rearchitecting` label. The PR explains why the tests live in the top-level `tests/` tree, which #332 leaves untouched, and I can move them to wherever the new layout wants them.

PR #335, opened 2026-08-10
tests: add integration tests for the decoding pipeline

mrviind:
## 🔗 Closed Issues

Towards #45, not closing it. That issue asks for the integration tests
Riverlane already has internally to be added to the repository. I have not
seen those tests, and three tests are a first slice of integration coverage,
not the whole of it. This PR adds newly written, CI-fast tests for the core
path (generate, simulate, decode, analyse, with real components) and leaves
#45 open for the rest.

---

## 📝 Description

Adds three integration tests that run the pipeline end to end with real
components and no mocks: circuit generation, noise compilation, Stim
simulation, matching decoding and analysis.

The tests use two physical error rates, both below the code's threshold,
each picked for what its test needs:

- p = 1e-4 makes logical failures rare, so a tight failure bound can prove
  error suppression is working.
- p = 1e-3 makes failures frequent enough to count reliably in a few
  thousand shots, which is what the determinism and distance-scaling tests
  need. It still sits safely below threshold: measured at p = 1e-2, the
  d=5 versus d=3 ordering reverses, so 1e-3 keeps a tenfold margin.

1. **Decoding sanity.** `run_decoding_on_circuit`, parametrised over
   d = 3, 5, 7 to match the `[3, 5, 7]` pattern used elsewhere in `tests/`,
   at p = 1e-4 with 200,000 shots per distance, must report fewer than 40
   failures. A working decoder gives about 8 failures at d=3 (measured 6 to
   13 across repeated runs) and near zero at d=5 and d=7. A decoder that
   ignores the syndromes gives about 1,000 at d=3, rising to about 2,400 at
   d=7, so the bound separates working from broken by two orders of
   magnitude at every distance.
2. **Determinism.** Two `StimDecoderManager` runs with the same seed must
   produce identical results at p = 1e-3. The run must also fail more than
   zero times and fewer than 100 (2% of 5,000 shots): a working decoder
   gives fewer than 20 failures here, a syndrome-ignoring one about 245, so
   the cap catches a broken decoder and the greater-than-zero floor catches
   a test that stopped seeing noise at all.
3. **Distance scaling.** At p = 1e-3, the logical error rate per round at
   d=5 must be lower than at d=3, and the d=3 value must sit under 3e-3,
   both computed with `compute_logical_error_per_round`. Measured: d=3
   gives about 1.5e-3 and d=5 about 1.4e-4, a tenfold suppression. The
   bounds come from measured failure distributions, including falsification
   runs with a syndrome-ignoring decoder and with scrambled detector
   wiring. Both are caught.

**Dependency floor (dropped).** An earlier version of this PR also raised
the pymatching floor in `deltakit-decode/pyproject.toml`, because these
were the first CI-run tests to reach `Matching.decode_batch` under the old
workspace resolution. #332 removed that file from this repository, so the
change was dropped in the merge commit. I will open it separately against
the deltakit-decode repository.

**Placement.** The tests use `deltakit_explorer`, `deltakit_circuit` and
`deltakit_decode` together, so they live in the top-level `tests/` tree
next to `tests/qmem/`. #332 kept that tree, and this repository now
installs deltakit-circuit, deltakit-core, deltakit-decode and
deltakit-explorer as regular dependencies, which is everything these tests
import. The standalone deltakit-decode repository's test dependencies do
not include `deltakit-explorer`, so the tests cannot run there as written.
I can move them if another location fits the new layout better; the file
is self-contained.

**Relation to `tests/qmem/`.** `test_qmem.py` runs a comparable real
pipeline, but it is `slow`-marked, so it does not run in CI, and it only
asserts that nothing raises. These tests run in a few seconds and assert
behaviour.

---

## 🚦 Status

Ready for review.

---

## 🛠️ Future Work

Two follow-ups, one on each side:

- #45's original request stays open: publishing the internal integration
  tests is work only the maintainers can do, and nothing here replaces it.
- On the contribution side, a negligible-noise variant of these tests for
  the repetition code would extend the same pattern to the second supported
  code family at little cost. I can pick that up as a follow-up PR if it
  would help.

---

## ➕️ Additional Information

**Verification** (all against `main` at 50ca5a3):

- Full suite with `uv sync --all-packages --group test` and
  `uv run --no-sync pytest`: 19,614 passed, 6 skipped, 3 failed in about
  4 minutes.
  The three failures are pre-existing image-comparison tests in
  `deltakit-explorer/tests/plotting` that fail in the same way on unmodified
  `main` in this environment (macOS matplotlib rendering against the
  committed reference images). They are unrelated to this change.
- New tests: 5 passed (3 parametrised distances plus 2) in 3 to 8 seconds
  per environment, from the command line and again in the PyCharm and
  VS Code test runners. The distance-scaling assertions were also checked
  across five seeds (1234, 7, 99, 2026, 424242) and hold on all five.
- CI matrix spot-checks, each from a fresh resolution with no lockfile,
  matching CI:
  Python 3.10 `highest` (pymatching 2.4.0, numpy 2.2.6): 5 passed.
  Python 3.13 `lowest-direct` (pymatching 2.2.2, numpy 2.1.0): 5 passed.
  Python 3.10 `lowest-direct` with the new floor (pymatching 2.2.2,
  numpy 2.0.0): 5 passed. The same cell without the floor change resolves
  pymatching 2.2.0 and all five cases fail in `Matching.decode_batch`, which
  is what motivated the dependency change.
- After #332 merged (`bcf5a70`): re-ran the five tests on the new `main`,
  with the subpackages installed from PyPI. 5 passed on the default
  resolution and 5 passed on `lowest-direct`, plus a manual run in the
  VS Code terminal.
- Static checks as the Pull Request workflow runs them: `ruff check`,
  `ruff format --check`, `pydoclint .` (the new file adds nothing to any
  baseline, checked under 0.8.x and 0.9.x), `typos`. Also clean: `bandit`
  with this repository's config (zero issues), `pip-audit` on the resolved
  environment (no known vulnerabilities), and the full pre-commit hook
  suite on both changed files.

**AI assistance**: Claude (Anthropic), used through Claude Code, drafted
these tests and this description. I directed and reviewed the work, read
every line of the diff, and ran the new tests myself in the PyCharm and
VS Code test runners before submitting. The main risk with drafted tests is
one that passes for the wrong reason, so the distance-scaling assertion was
re-run across the five seeds listed above rather than accepted from a single
run, and the physical error rate it uses sits below threshold on purpose: at
or above threshold a larger distance gives a worse logical error rate, and
the same assertion would fail for a correct QEC code.

**Licence**: this contribution contains no third-party material. The tests
call only this repository's own public API and follow the structure of the
existing suites here. I read the full diff for anything carried in from
elsewhere and there is none, so nothing here needs the CLA's "Submission
containing materials of a third-party:" marking. On the AI-drafted part: I
use Claude under a paid subscription, and Anthropic's Consumer Terms of
Service (section 4, "Inputs, Outputs, Actions, and Materials") state that
Anthropic assigns to the user all of its right, title and interest, if any,
in Outputs; those terms cover Claude Code used from consumer accounts. The
drafted content is therefore mine to submit. Reference:
https://www.anthropic.com/legal/consumer-terms. I have signed the CLA and
submit this contribution under Apache 2.0.

---

## 🧾 Release Note

tests: add integration tests for the decoding pipeline


rolandriver:
Hi @aravind3 thank you for your contribution to Deltakit ! Unfortunately, we can't follow-up on this PR now as per #332 subpackages will be removed from `deltakit` root. Can I ask you to move it to the relevant [deltakit-decode repo](https://github.com/Deltakit/deltakit-decode)? And we will take it from there. Thank you so much.

mrviind:
Thanks @rolandriver. Quick note before I move anything, these tests are in the top-level `tests/` tree, not a subpackage, and #332 kept that tree. I re-ran them on the new `main` just now. 5 passed (the repo still installs all four packages, which is all they need).

The only subpackage piece here was the `deltakit-decode/pyproject.toml` change, which #332 has now deleted. I have dropped it, resolved the conflict, retitled the PR and updated the description to match: it is now the test file only, and I will raise the pymatching floor separately against the deltakit-decode repo.

If you would still prefer the tests themselves in the deltakit-decode repo, I can make them decode-self-contained with stim's built-in generator and reopen there. Whichever you prefer, I will move forward on that direction.

nelimee:
After the split into independent sub-modules, we don't have a place for integration tests any more. That's something we need to fix. Initially, we wanted this repository `deltakit` (which was previously the mono-repository) to host only the documentation. It might be the correct place for such integration tests, but I would like to discuss it more before taking that decision.

Don't expect an answer before early September. I'll review the PR by then, once we have an answer on where this code fits best.

mrviind:
Thanks @nelimee, understood. The file is self-contained so it can move to whichever repository you settle on. I will leave the PR as it is until you come back to it.

rolandriver (approved):
Thank you @mrviind for your contribution to Deltakit! This is looking good to me, a general comment about adding type hints to variables. I would also wait for @nelimee's approval.

tests/integration/test_decode_pipeline_integration.py
```diff
@@ -0,0 +1,138 @@
+# (c) Copyright Riverlane 2020-2025.
+"""Three tests that run the circuit-generation, simulation, decoding and
+analysis pipeline end to end with real components rather than mocks.
+
+A first slice towards https://github.com/Deltakit/deltakit/issues/45.
```
nelimee: Should be removed
```suggestion
```

